PR #1, opened 2017-02-09
profile.getEmailAddress() always returns null

markmclaren:
Fixed problem when null is always returned from profile.getEmailAddress(). 

Mapping attribute name in HumanProfileAdapter is wrong.


qzagarese:
Thanks Mark! Really appreciated :-)

PR #4, opened 2017-03-20
Added Spring boot auto configuration module

davidgoate:
Added Spring boot auto configuration module to simplify integration of the Java SDK with Spring Boot based applications. 

- Updated the README.
- Updated the spring boot sample application to use the new library.

### DIFF
--- a/java-sdk-spring-boot-auto-config/pom.xml
+++ b/java-sdk-spring-boot-auto-config/pom.xml
@@ -60,13 +60,13 @@
     <dependency>
       <groupId>com.yoti</groupId>
       <artifactId>java-sdk-api</artifactId>
-      <version>1.0-SNAPSHOT</version>
+      <version>1.0</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>com.yoti</groupId>
       <artifactId>java-sdk-impl</artifactId>
-      <version>1.0-SNAPSHOT</version>
+      <version>1.0</version>
       <scope>provided</scope>
     </dependency>
 
@@ -204,14 +204,6 @@
       </plugin>
     </plugins>
   </build>
-
-  <repositories>
-    <repository>
-      <id>yoti-sdk</id>
-      <name>libs-snapshot-local</name>
-      <url>https://www.yoti.com/downloads/sdk/libs-snapshot-local/</url>
-    </repository>
-  </repositories>
 
   <distributionManagement>
     <snapshotRepository>
--- a/java-sdk-spring-boot-auto-config/pom.xml
+++ b/java-sdk-spring-boot-auto-config/pom.xml
@@ -1,0 +1,227 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <properties>
+    <maven.compiler.source>1.6</maven.compiler.source>
+    <maven.compiler.target>1.6</maven.compiler.target>
+    <owasp.dependency.check.cvss.limit>4</owasp.dependency.check.cvss.limit>
+    <owasp.dependency.check.cve.update.limit>24</owasp.dependency.check.cve.update.limit>
+    <findbugs.version>3.0.0</findbugs.version>
+  </properties>
+
+  <groupId>com.yoti</groupId>
+  <artifactId>java-sdk-spring-boot-auto-config</artifactId>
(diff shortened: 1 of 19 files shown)


davidgoate:
Note: I have set the pom version of the new module to 1.0 but it will need publishing into sonatype for the dependency to be available to clients.

java-sdk-spring-boot-auto-config/pom.xml
```diff
@@ -0,0 +1,227 @@
...
+    <dependency>
+      <groupId>com.yoti</groupId>
+      <artifactId>java-sdk-api</artifactId>
+      <version>1.0-SNAPSHOT</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.yoti</groupId>
+      <artifactId>java-sdk-impl</artifactId>
+      <version>1.0-SNAPSHOT</version>
```
qzagarese: Please rely on the stable release from Central.

java-sdk-spring-boot-auto-config/pom.xml
```diff
@@ -0,0 +1,227 @@
...
+        </dependencies>
+      </plugin>
+    </plugins>
+  </build>
+
+  <repositories>
+    <repository>
+      <id>yoti-sdk</id>
+      <name>libs-snapshot-local</name>
+      <url>https://www.yoti.com/downloads/sdk/libs-snapshot-local/</url>
```
qzagarese: This repository is not needed any more.

java-sdk-spring-boot-auto-config/src/main/java/com/yoti/api/spring/YotiClientProperties.java
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Allows properties to the Yoti client to be supplied via spring properties (e.g. YAML or properties file).
+ */
+@ConfigurationProperties(prefix = "com.yoti.client")
+public class YotiClientProperties {
+
+    /**
+     * The SDK client id provided by Yoti dashboard.
+     */
+    private String clientSdkId;
```
qzagarese: should SDK be all uppercase, given it is an acronym? It is not a rhetorical question.

qzagarese:
CR done.

java-sdk-spring-boot-auto-config/src/main/java/com/yoti/api/spring/YotiClientProperties.java
davidgoate: @qzagarese  @gitplaneta This would seem to be a somewhat unclear and potentially contentious issue. I was also not clear on this and from what I have been able to established with a Google search it seems that there is actually no official guideline from Sun/Oracle on this for the Java language. Most recommendations seem to be opinion based but one thing that does seem to be clear is that consistency is important so if we adopt a rule then it should be applied consistently.

- [Official guidelines](http://www.oracle.com/technetwork/java/codeconventions-135099.html) (albeit from 1999) don't seem to indicate what should be done with acronyms within variable names. 
- [PMP name rules](https://pmd.github.io/pmd-5.0.0/rules/java/naming.html#VariableNamingConventions) also do not take a stance on this.
- Some Stack Overflow questions on this subject have different opinions but all seem to advocate consistency. E.g like [this](http://stackoverflow.com/questions/2236807/java-naming-convention-with-acronyms) or [this](http://stackoverflow.com/questions/8277355/naming-convention-for-upper-case-abbreviations) or [this](http://stackoverflow.com/questions/414001/variable-naming-conventions-in-java).


One observation I'd make here is that could the same argument be applied to "Id" if is considered an abbreviation of Identifier? E.g. could be we saying that variables with the word `Id` should be `ID`.  And if so, for readability would we consider that the variable name could be one of: `clientSDKId`, `clientSdkId` or `clientSDKID`.

More food for thought/philosophical I presume. For me I'd consider the latter less readable than either of the first two options, but perhaps others would disagree.


Another observation would be that in other areas of our Java SDK source-base we have used lowercase for acronyms such as "URI", for example in [the Receipt class](https://github.com/getyoti/java/blob/master/java-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/Receipt.java#L25) or for [URL](https://github.com/getyoti/java/blob/master/java-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/RemoteProfileService.java#L45)

My personal feelings on this is that as far as I know there isn't a clear official Sun/Oracle guideline and that there also doesn't appear to be an obvious de-facto standard either so it's just important for us to decide as a team what the right approach for our project is and apply this rule consistently. I would not be upset with either approach. If I were forced to take a position and not to "sit on the fence" I'd personally leave this variable name as it is. But equally, I would understand where you were coming from if you decide that it ought to be upper cased.

qzagarese: Thank you, that was quite an extensive explanation. I'm good with keeping it as is.

PR #5, opened 2017-03-26
Allow HTTP read/connect timeouts to be configured for the Yoti Client

davidgoate:
For a client application calling the `getActivity` method of the `YotiClient` could be considered an unsafe/dangerous operation since you'd give up control of your application thread to our 3rd party code which internally has the potential to indefinitely block. This could cause thread pools on the client to all become blocked in a waiting state, eventually brining the client down (e.g. if all server worker threads are waiting). 

This change provides two system properties to allow the client HTTP connection timeout behaviour to be configured.

* `yoti.client.connect.timeout.ms` - the number of milliseconds that you are prepared to wait for the connection to be established. Zero is interpreted as an infinite timeout.
* `yoti.client.read.timeout.ms` - the number of milliseconds that you are prepared to wait for data to become available to read in the response stream. Zero is interpreted as an infinite timeout.

I have also updated the read me docs to reflect this change and removed a reference to the`activityDetails.getOutcome()` method since this appears to have been removed from the API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,17 +75,13 @@ Here is an example of how this works:
 ```java
 ActivityDetails activityDetails;
 try {
-	activityDetails = client.getActivityDetails(token);
-    if(activityDetails.getOutcome().isSuccessful()) {
-    	Optional<YourAppUserClass> user = yourUserSearchMethod(activityDetails.getUserId());
-    	if(user.isPresent()) {
-    		// handle login
-    	} else {
-    		// handle registration
-    	}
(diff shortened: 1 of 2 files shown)


qzagarese:
Thank you @davidgoate ! This will be merged soon. Need an hour to be able to release.

davidgoate:
Closing as this branch was merged into https://github.com/getyoti/java/pull/10 and will be merged there instead.

This PR will *not* be merged.

PR #8, opened 2017-06-08
Documentation update

davidgoate:
- Make it more clear that tokens are single use only.
- Clean up example code after removal of the sharing outcome object.


davidgoate:
@qzagarese Intended to help prevent similar mistake to MyVenue problem.

PR #13, opened 2017-08-22
Maven Wrapper, Security Scanner & Documentation Updates

davidgoate:
In summary:

## Documentation
- Added imports and package information to the README to make it clearer (as recommended by @vassyz  in a previous issue he raised).
- Fixed some numbering/formatting in the README .
- Added info on how to skip certain build tools like security checks.

## Maven Wrapper (Build Consistency)
- Much like `gradlew`.
- Added the maven wrapper (./mvnw) at version 3.3.9 to make the build more consistent (even for users without any maven installed or a different version locally).

## POM House Keeping

* Added/modified the POM `<name>` for some modules that were not clear or were missing one for a more complete/friendly build reactor.
* Fixed minor issue with missing plugin version in POM that causes build warnings.

## Static Security Analysis

- Added findbugs security extensions (http://find-sec-bugs.github.io/)
- This can be skipped during the build with user property `findbugs.skip` e.g. `mvn clean install -Dfindbugs.skip=true`
- Added exclusions for cipher without integrity check (SecureYotiClient) for using `AES/CBC/PKCS7Padding` where `AES-GCM` or `AES-CCM` would have been better but cannot be supported on Android clients older than API level 26 (https://developer.android.com/reference/javax/crypto/Cipher.html). Also now, the server and all other clients could require changing. Checked this with laurence.withers@yoti.com, henry.coggill@yoti.com and quirino.zagarese@yoti.com. Risk seems to be acceptable so excluding warning.
- Also excluding all warnings for CRLF log injection. If a client app wants to mitigate this they can by configuring their underlying logger to strip CRLF, e.g. logback: `<pattern>%-5level - %replace(%msg){'[\r\n]', ''}%n</pattern>`.

## Urgency

No hurry for any of these. Since it is all documentation and build changes (no actual code) a new version does **not** need to be published to Maven Central if/after this is merged.


davidgoate:
@qzagarese I think they are ready.

davidgoate:
@qzagarese actually, shall we wait for the outcome of the testing on the encoding of PEM's as that might require additional work on the readme to explain how to re-encode on 1.1 if necessary.

qzagarese:
Up to you @davidgoate .
For me it is already good to merge this.

davidgoate:
@qzagarese this is now ready with additional section added to describe recently discovered PEM issues since upgrade.

PR #19, opened 2017-10-13
[SDK-118]: Added Yoti-SDK to request headers

jguthrie100:


vassyz:
@qzagarese  We're adding the `X-Yoti-SDK` header value to all the SDKs to get a better understanding of the language usage.

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/call/RemoteProfileServiceTest.java
```diff
@@ -86,6 +87,10 @@ private void assertAuthKey(KeyPair keyPair, Map<String, String> headers) {
         assertEquals(base64(keyPair.getPublic().getEncoded()), headers.get("X-Yoti-Auth-Key"));
     }
 
+    private void assertYotiSDK(KeyPair keyPair, Map<String, String> headers) {
+        assertEquals("Java", headers.get("X-Yoti-SDK"));
```
qzagarese: this can be extracted to a constant

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/call/RemoteProfileServiceTest.java
```diff
@@ -32,6 +32,7 @@
 
 public class RemoteProfileServiceTest {
     private static final String MESSAGE_PREFIX = "GET&";
+    private static final String YOTI_SDK_HEADER = "X-Yoti-SDK";
```
vassyz: Aren't we calling it SDK identifier in the other SDKs?

echarrod: I just used the same name we have in RemoteProfileService.java, where we use:
```Java
private static final String AUTH_KEY_HEADER = "X-Yoti-Auth-Key";
private static final String DIGEST_HEADER = "X-Yoti-Auth-Digest";
private static final String YOTI_SDK_HEADER = "X-Yoti-SDK";
````

vassyz: Ok

PR #32, opened 2017-12-20
[SDK-190]: Updated README with extra command to use when running the …

echarrod:
…example project with Java 9

Wasn't entirely sure about the wording - or whether to mention anything when we instruct the users to run the original command `java -jar target/yoti-sdk-spring-boot-example-1.3.jar` above.

### DIFF
--- a/yoti-sdk-spring-boot-example/README.md
+++ b/yoti-sdk-spring-boot-example/README.md
@@ -29,13 +29,13 @@ Before you start, you'll need to create an Application in [Dashboard](https://ww
 
 ## Running
 * You can run your server-app by executing `java -jar target/yoti-sdk-spring-boot-example-1.3.jar`
+  * If you are using Java 9, you can run the server-app as follows `java -jar target/yoti-sdk-spring-boot-example-1.3.jar --add-exports java.base/jdk.internal.ref=ALL-UNNAMED`
 * Your endpoint is listening under `https://localhost:8443/login`.
 
 In order to receive calls on your /login endpoint, you need to expose your server-app to the outside world. We require that you use the domain from the Callback URL and HTTPS.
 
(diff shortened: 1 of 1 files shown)


yoti-sdk-spring-boot-example/README.md
```diff
@@ -35,6 +35,7 @@ In order to receive calls on your /login endpoint, you need to expose your serve
 
 ## Requirements for running the application
 * Java 6 or above
+* If you are using Java 9, you should add an `add-exports` argument to the command you use to run the .jar file, so that it is `java -jar target/yoti-sdk-spring-boot-example-1.3.jar --add-exports java.base/jdk.internal.ref=ALL-UNNAMED`
```
qzagarese: How about putting this in the "Running"  section?
You could use a sub item in the list as explained here https://help.github.com/articles/basic-writing-and-formatting-syntax/#lists

In terms of wording, something like "If you are using Java 9, you can run the server-app as follows"  `java -jar target/yoti-sdk-spring-boot-example-1.3.jar --add-exports java.base/jdk.internal.ref=ALL-UNNAMED`

echarrod:
[How it looks in the README](https://github.com/getyoti/yoti-java-sdk/tree/Java9/yoti-sdk-spring-boot-example#running)

vassyz:
Looks alright to me, but I think the whole page needs to be checked by a copyrighter. Should the header `Building and running your example server-app` mention `and running`? Seems we are addressing the running part in the next section.

echarrod:
@vassyz yes good point - it should either be `Building and Running your example server-app`, with subsections for `Building` and `Running`, but I think it would be better if we just removed the `and running` from the 1st title. It would be good to get all the READMEs copied by a copywriter, what's the process for doing that?

PR #34, opened 2018-01-03
FIX-BUILD Update module versions

bucky-boy:
Even after this change, you still need to run with the -Denforcer.skip=true flag


qzagarese:
@echarrod @vassyz the previous master could not be built. My bad

PR #35, opened 2018-01-04
[SDK-218]: Fixed inter-project dependencies

echarrod:
so the -Denforcer.skip flag doesn't have to be specified on build


DarthSeppius:
seems good for me, LGTM.

PR #37, opened 2018-01-25
Api 264

bucky-boy:
Francesco - There's a fair few changes here now, in a number of commits.  I've tested it against staging, it works.
Quirino - for info


yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/factory/SignatureFactory.java
```diff
@@ -1,26 +1,42 @@
 package com.yoti.api.client.spi.remote.call.factory;
 
-import com.yoti.api.client.ProfileException;
-
 import java.security.GeneralSecurityException;
 import java.security.PrivateKey;
 import java.security.SecureRandom;
 import java.security.Signature;
 
+import static com.yoti.api.client.spi.remote.Base64.base64;
```
DarthSeppius: put this on top pls

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/factory/MessageFactory.java
```diff
@@ -0,0 +1,23 @@
+package com.yoti.api.client.spi.remote.call.factory;
+
+import java.io.UnsupportedEncodingException;
+
+import static com.yoti.api.client.spi.remote.Base64.base64;
+import static com.yoti.api.client.spi.remote.call.YotiConstants.DEFAULT_CHARSET;
+
+public class MessageFactory {
```
DarthSeppius: This class seems not to be a Factory class if we mean `Factory` as the Design Pattern. Maybe we can use another name, like MessageGenerator for example.

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/factory/SignatureFactory.java
```diff
@@ -1,26 +1,42 @@
...
+
 public class SignatureFactory {
```
DarthSeppius: Same comment as for the MessageFactory.

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/factory/SignatureFactory.java
```diff
@@ -1,26 +1,42 @@
...
+    public String create(PrivateKey privateKey, String httpMethod, String path, byte[] body) throws GeneralSecurityException {
+        byte[] message = messageFactory.create(httpMethod, path, body);
+        return createSignature(message, privateKey);
+    }
+
+    private String createSignature(byte[] message, PrivateKey privateKey) throws GeneralSecurityException {
+        Signature signature = Signature.getInstance(SIGNATURE_ALGORITHM, BOUNCY_CASTLE_PROVIDER);
+        signature.initSign(privateKey, new SecureRandom());
+        signature.update(message);
+        return base64(signature.sign());
```
DarthSeppius: the method is called `createSignature` but actually, it creates a Signature obj and also signs it, generating a string that is the message signed. If the name of the class is SignatureFactory and we are thinking of having a different implementation of the `java.security.Signature` obj, it's a good idea to have it as a parameter during the init of the SignatureFactory class.

My suggestion is that we can return the Signature obj or if we want to sign also the message we can call the method in a different way, like `signMessage`. Also keep in mind that when we call `sign()` the state of the Signature object will be reverted to ` signature.initSign(privateKey, new SecureRandom());`. You can use the same instance for signing multiple times, there is no need to create a new one every time.

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/call/aml/RemoteAmlServiceTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+        SimpleAmlResult result = testObj.performCheck(keyPair, SOME_APP_ID, amlProfileMock);
+
+        verify(resourceFetcherMock).postResource(urlConnectorCaptor.capture(), eq(SERIALIZED_BODY.getBytes()), headersCaptor.capture(), eq(SimpleAmlResult.class));
+        assertUrl(urlConnectorCaptor.getValue());
+        assertHeaders(headersCaptor.getValue());
+        assertSame(result, simpleAmlResultMock);
+    }
+
+    private void assertUrl(UrlConnector urlConnector) throws MalformedURLException {
```
DarthSeppius: I not sure that is worth to have all this private method. I think it's better to put the code inside the `shouldPerformAmlCheck()`. And if the problem is the clarity of the checks then it's better to have test variables with more expressive names, like SOME_SIGNATURE ~ TEST_DIGEST and so on. I know that is the same approach or similar of the other test but seems this is a new one I think that we can think about that at least in this class.

yoti-sdk-dummy/src/main/java/com/yoti/api/client/spi/dummy/DummyYotiClient.java
```diff
@@ -195,6 +203,12 @@ public ActivityDetails getActivityDetails(String encryptedConnectToken) throws P
         return DUMMY_RECEIPT;
     }
 
+    //FIXME - Do not check in like this
```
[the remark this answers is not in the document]
bucky-boy: Yes.  We need to clarify what we're doing with the dummy project.  I think at some point Quirino said we should delete it.

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/factory/MessageFactory.java
bucky-boy: To me, it creates Messages, so it's a MesageFactory.
It is definitely doing some creation.  In this case it returns an Array object instead of a Message object, but MessageFactory is probably a nicer name than ArrayFactory or MessageArrayFactory.

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/factory/SignatureFactory.java
bucky-boy: > the method is called createSignature but actually, it creates a Signature obj and also signs it, generating a string that is the message signed.
Signing the signature is fine, that's just a part of our creation process.  The real problem then is returning a String rather than a Signature.  It takes byte[] message, so maybe it should be called signMessage.
> it's a good idea to have it as a parameter during the init of the SignatureFactory class
I doubt we'll need to use a different implementation of Signature.  That might be more than we need now.

> My suggestion is that we can return the Signature obj or if we want to sign also the message we can call the method in a different way, like signMessage
So the private method could be signMessage, and we could call the class SignedMessageFactory.  Would that be more appropriate?

> You can use the same instance for signing multiple times, there is no need to create a new one every time
If we do that, don't we need to synchronize on it?

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/call/aml/RemoteAmlServiceTest.java
bucky-boy: > I think it's better to put the code inside the shouldPerformAmlCheck()

I tried it inline, it was awful.

> And if the problem is the clarity of the checks then it's better to have test variables with more expressive names

I don't think the variable names are the problem.  It's the boiler plate around reaching what we want to test that is causing a mess.  Maybe hamcrest would clean it up.

bucky-boy: hamcrest was too much.  See if its better now.

PR #39, opened 2018-02-07
API-264: Fix creation of AML beans, to match approach take elsewhere

bucky-boy:
This commit changes how AML are created.  Now clients can use builder, and the only have to reference classes from the sdk-api project in their code (though they do have to import the sdk-impl project anyway).

### DIFF
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/aml/AmlAddressFactory.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/aml/AmlAddressFactory.java
@@ -1,0 +1,7 @@
+package com.yoti.api.client.aml;
+
+public interface AmlAddressFactory {
+
+    AmlAddress create(String postCode, String country);
+
+}
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/aml/AmlProfileBuilder.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/aml/AmlProfileBuilder.java
@@ -1,0 +1,41 @@
+package com.yoti.api.client.aml;
+
+import java.util.ServiceLoader;
+
+public class AmlProfileBuilder {
+
+    private String givenNames;
+    private String familyName;
+    private String ssn;
+    private AmlAddress amlAddress;
+
+    public AmlProfileBuilder withGivenNames(String givenNames) {
+        this.givenNames = givenNames;
+        return this;
+    }
+
+    public AmlProfileBuilder withFamilyName(String familyName) {
+        this.familyName = familyName;
+        return this;
+    }
+
+    public AmlProfileBuilder withSsn(String ssn) {
+        this.ssn = ssn;
+        return this;
+    }
+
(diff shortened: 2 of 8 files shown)


yoti-sdk-api/src/main/java/com/yoti/api/client/aml/AmlAddressFactory.java
```diff
@@ -0,0 +1,7 @@
+package com.yoti.api.client.aml;
+
+public interface AmlAddressFactory {
+
+    AmlAddress create(String postCode, String country);
+
```
qzagarese: Do we need one with country only too?

yoti-sdk-api/src/main/java/com/yoti/api/client/aml/AmlProfileBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+        this.givenNames = givenNames;
+        return this;
+    }
+
+    public AmlProfileBuilder withFamilyName(String familyName) {
+        this.familyName = familyName;
+        return this;
+    }
+
+    public AmlProfileBuilder withSsn(String ssn) {
```
qzagarese: should SSN be uppercase?

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/aml/SimpleAmlAddressFactory.java
```diff
@@ -0,0 +1,12 @@
+package com.yoti.api.client.spi.remote.call.aml;
+
+import com.yoti.api.client.aml.AmlAddress;
+import com.yoti.api.client.aml.AmlAddressFactory;
+
+public class SimpleAmlAddressFactory implements AmlAddressFactory {
+
+    public AmlAddress create(String postCode, String country) {
```
qzagarese: do we need another one with country only?

yoti-sdk-api/src/main/java/com/yoti/api/client/aml/AmlAddressFactory.java
bucky-boy: No, because there is a builder now.  They can do:
```
AmlAddress amlAddress = new AmlAddressBuilder()
      .withCountry("xyz")
      .build();
```

yoti-sdk-api/src/main/java/com/yoti/api/client/aml/AmlProfileBuilder.java
bucky-boy: I don't think so, it's a variable name

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/aml/SimpleAmlAddressFactory.java
bucky-boy: Same as the other answer - I assume people will use the builder

PR #43, opened 2018-02-08
API-357: Add age verification helper method to HumanProfile

bucky-boy:
Here's the change to add isAgeVerified.
I've also taken this opportunity to update bouncy-castle to latest (1.59)

### DIFF
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/HumanProfile.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/HumanProfile.java
@@ -6,6 +6,7 @@ import java.util.Map;
  * Profile of an human user with convenience methods to access well-known attributes.
  */
 public interface HumanProfile extends Profile {
+
     /**
      * Corresponds to primary name in passport, and surname in English.
      *
@@ -111,4 +112,5 @@ public interface HumanProfile extends Profile {
     static enum Gender {
         MALE, FEMALE, OTHER
     }
+
 }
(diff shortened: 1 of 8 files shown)


yoti-sdk-api/src/main/java/com/yoti/api/client/HumanProfile.java
```diff
@@ -44,6 +44,13 @@
      */
     Date getDateOfBirth();
 
+    /**
+     * Did the user pass the age verification check?
+     *
+     * @return <code>TRUE</code> if they passed, <code>FALSE</code> if they failed, <code>null</code> if there was no check
+     */
+    Boolean isAgeVerified();
+
```
DarthSeppius: only one question, we need to provide also the implemented one in the HumanProfileAdapter, right?

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/HumanProfileAdapter.java
```diff
@@ -87,6 +96,13 @@ public Date getDateOfBirth() {
         return wrapped.getAttribute(ATTRIBUTE_DOB, Date.class);
     }
 
+    @Override
+    public Boolean isAgeVerified() {
+        Boolean isAgeOver = wrapped.getAttribute(ATTRIBUTE_AGE_OVER, Boolean.class);
```
qzagarese: For what I can see, this cannot work.
The attribute starts with `age_over:` or `age_under:` (e.g. `age_over:18`)

I believe you need changes in the implementation of Profile where you can query for an `age_under:*` or `age_under:*` and return one if available.

DarthSeppius: Yep, see on this page `https://lampkicking.atlassian.net/wiki/spaces/IPASS/pages/29556930/Attribute+and+backing+document+names`

bucky-boy: Fixed

PR #46, opened 2018-02-14
SDK-264: Added getFullName, deprecated getGivenAndLastNames

echarrod:
Didn't make any changes to HumanProfileAdapterTests in the end, as there is no logic in HumanProfileAdapter for GetFullName, it simply returns the attribute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ Yoti will provide a boolean result on the following checks:
 * Fraud list - Verify against  US Social Security Administration Fraud (SSN Fraud) list
 * Watch list - Verify against watch lists from the Office of Foreign Assets Control
 
-To use this functionality you must ensure your application is assigned to your Organisation in the Yoti Dashboard - please see here for further information.
+To use this functionality you must ensure your application is assigned to your Organisation in the Yoti Dashboard - please see [here](https://www.yoti.com/developers/documentation) for further information.
 
 For the AML check you will need to provide the following:
 
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Description on handling user logons
 Description of network connectivity requirements
 
 1) [AML Integration](#aml-integration) -
-Integrating with Yoti's AML service
+How to integrate with Yoti's AML (Anti Money Laundering) service
 
 1) [Modules](#modules) -
 The Modules above explained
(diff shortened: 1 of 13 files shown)


yoti-sdk-api/src/main/java/com/yoti/api/client/HumanProfile.java
```diff
@@ -24,18 +24,18 @@
...
      */
-    @Deprecated
     String getFullName();
 
     /**
      * Equal to ${given_names} + " " + ${family_name}.
      *
      * @return the given names + the surname
      * @since 1.2
+     * @deprecated please use {@link #getFullName() instead}
```
qzagarese: I don't think we should suggest to use getFullName instead of this.
This one is a convenience method that returns the concatenation of given_names and family_name.
`getFullName` returns the full_name attribute.

If the qrcode requests given_names and family_name, full_name will not be there.
Similarly, if the qr code requests full_name, then given_names and family_name will not be there.

The reason for having these 3 attributes (and specifically full_name ) is that certain ID documents like the Indian Aadhaar do not provide the name components, but only the full_name.

echarrod: Ah, I thought that would `GetFullName` would return the equivalent of `GetGivenAndLastName`, if only the `Given Name` and `Family Name` are present (and not `FullName`). 

If that's not the case, do we still want to deprecate `GetGivenAndLastName`, if it's providing us with a helper method which we're not getting otherwise?

qzagarese: We don't really need to deprecate it although I don't how much value it actually adds.
@bucky-boy  thoughts?

bucky-boy: I think we need to deprecate it, it was my suggestion.
We now have getFullName, which currently describes itself as "${given_names} + " " + ${family_name}" as well as getGivenAndLastNames.  I'd suggest we try to keep it trimmed down

yoti-sdk-api/src/main/java/com/yoti/api/client/HumanProfile.java
```diff
@@ -24,18 +24,18 @@
     /**
      * Equal to ${given_names} + " " + ${family_name}.
```
bucky-boy: I'm not sure this statement is correct.
Sorry for being pedantic, but I think this is quite important.  We need to nail these definitions, client devs will take them as gospel, and this kind of attribute is right at the core of what we do.

echarrod: This is what I assumed originally, realised it was wrong after Quirino pointed it out above, I'll change it.

yoti-sdk-api/src/main/java/com/yoti/api/client/HumanProfile.java
echarrod: I agree, I think we can deprecate getGivenAndLastNames, making it simple, and if the client wants to add a similar method they can do this easily.

README.md
```diff
@@ -206,38 +207,38 @@ Since version 1.1 of the `yoti-sdk-impl` you can set the following two system pr
...
 Yoti will provide a boolean result on the following checks:
+
 * PEP list - Verify against Politically Exposed Persons list
 * Fraud list - Verify against  US Social Security Administration Fraud (SSN Fraud) list
 * Watch list - Verify against watch lists from the Office of Foreign Assets Control
 
-To use this functionality you must ensure:
-* Your application is assigned to your Organisation in the Yoti Dashboard - please see [here](https://www.yoti.com/developers/documentation) for further information.
-* Within your application please ensure that you have selected the 'given names' and 'family name' attributes from the data tab. This is the minimum requirement for the AML check.
+To use this functionality you must ensure your application is assigned to your Organisation in the Yoti Dashboard - please see here for further information.
```
bucky-boy: Missing a link for 'here'

bucky-boy:
@echarrod - I've bumped the version to 1.4.1-SNAPSHOT.  Review that change, and when you are happy you can merge it

PR #50, opened 2018-03-05
API-413: Update README for isAgeVerified() bug

bucky-boy:
Is this a reasonable update for isAgeVerified()?

I'll bump the version numbers to 1.4.2 when it has been released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ If you are using Maven, you need to add the following dependency:
 
 If you are using Gradle, here is the dependency to add:
 
-`compile group: 'com.yoti', name: 'yoti-sdk-impl', version: '1.3'`
+`compile group: 'com.yoti', name: 'yoti-sdk-impl', version: '1.4.1'`
 
 You will find all classes packaged under `com.yoti.api`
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -102,7 +102,7 @@ If you are using Maven, you need to add the following dependency:
 
 If you are using Gradle, here is the dependency to add:
 
-`compile group: 'com.yoti', name: 'yoti-sdk-impl', version: '1.3'`
+`compile group: 'com.yoti', name: 'yoti-sdk-impl', version: '1.4.1'`
```
qzagarese: we are re-updating this once we release the fix, right?

README.md
```diff
@@ -302,6 +303,20 @@ Combining this with the Spring Boot Auto Configuration can make integration very
...
 
+### Age Verification
+
+#### Affects
+
+* Versions prior to 1.4.2
+
+#### Description
+
+A bug in the isAgeVerified() property of the profile meant that isAgeVerified() would always return null (i.e. not checked).
```
vassyz: Could you please format the method name like this `isAgeVerified()`?

vassyz:
Looks good to me. Is it possible to add `isAgeVerified` to the example?

README.md
bucky-boy: Yes.  As soon as 1.4.2 is out there, I will update the readme.

README.md
bucky-boy: Sure

bucky-boy:
If you mean add it to the spring example, that's a good idea.  For now I'd rather get this fix deployed, then add that in when we get an opportunity to test it (that is, test the spring example too).

PR #51, opened 2018-03-08
API-407: Support for extended document_details attribute in Java SDK

bucky-boy:
There was a problem where DocumentDetails would not be parsed if no optional attributes were provided, now fixed.  It now also supports the new _issuingAuthority_ attribute.

For now, the new document types have been included in the enum because removing the enum would be a breaking change.  There's a more detailed discussion in the ticket.
https://lampkicking.atlassian.net/projects/API/issues/API-407

This doesn't have to go in, in its present form.  If we review the ticket and decide to do something about the questions raised then we'll address them first before merging/releasing this.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <groupId>com.yoti</groupId>
   <artifactId>yoti-sdk</artifactId>
   <packaging>pom</packaging>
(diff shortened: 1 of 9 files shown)


bucky-boy:
@qzagarese says this is 'approved' so I'm merging it

PR #55, opened 2018-04-18
SDK-319 Address formatting nitpicks

bucky-boy:


yoti-sdk-spring-boot-auto-config/src/main/java/com/yoti/api/spring/YotiClientProperties.java
```diff
@@ -84,10 +84,9 @@ public boolean equals(Object o) {
         if (o == null || getClass() != o.getClass()) return false;
 
         YotiClientProperties that = (YotiClientProperties) o;
-
-        if (applicationId != null ? !applicationId.equals(that.applicationId) : that.applicationId != null)
-            return false;
+        if (applicationId != null ? !applicationId.equals(that.applicationId) : that.applicationId != null) return false;
```
irotech: we'd use always { } even for just one line

PR #57, opened 2018-04-19
API-476: Final tidy up

bucky-boy:
Some changes I chose to make, some were forced upon me....

**CVE**
There were some more CVE errors, this time for Spring.  One is from spring-boot, one is spring-core (security).  Rather than add more CVE exclusions,  I have bumped the versions of both to the lowest possible release needed.

**finbugs**
There were a few:
1. It didn't like handling _Exception_
2. It didn't like deserialisation of an object.  It turns out we don't have to use _InputStream_ after all.  Since we have a String, we can just use its bytes and get the BC _ASN1Primitive_ that way.  The tests pass, so I believe it works.

### DIFF
--- a/yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/AttributeConverter.java
+++ b/yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/AttributeConverter.java
@@ -8,69 +8,53 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.yoti.api.client.Attribute;
 import com.yoti.api.client.spi.remote.proto.AttrProto;
 import com.yoti.api.client.spi.remote.proto.AttrProto.Anchor;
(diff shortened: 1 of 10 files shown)


yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/AttributeConverter.java
```diff
@@ -25,52 +25,37 @@
...
+                return attributeWithSources(attribute, attribute.getValue().toString(DEFAULT_CHARSET));
         }
+    }
 
-        LOG.error("Unknown type {} for attribute {}", attribute.getContentType(), attribute.getName());
-        return new com.yoti.api.client.Attribute(attribute.getName(), 
-                attribute.getValue().toString(DEFAULT_CHARSET),
-                extractMetadata(attribute, AnchorType.SOURCE),
-                extractMetadata(attribute, AnchorType.VERIFIER));
+    private static Attribute attributeWithSources(AttrProto.Attribute attribute, Object value) {
```
qzagarese: how about `attributesWithMetadata` ?

bucky-boy: Done

PR #59, opened 2018-06-07
API-554: Use a parent pom for common configuration

bucky-boy:
I've made some significant changes to the build.  Please see API-554, I've described the changes there


yoti-sdk-parent/pom.xml
```diff
@@ -0,0 +1,419 @@
...
+    
+    <springboot.version>1.5.10.RELEASE</springboot.version>
+    <spring.core.version>4.3.14.RELEASE</spring.core.version>
+    <spring.security.version>4.2.3.RELEASE</spring.security.version>
+    <javax.servlet.version>2.5</javax.servlet.version>
+    
+    <junit.version>4.12</junit.version>
+    <mockito.version>2.8.47</mockito.version>
+    <hamcrest.version>1.3</hamcrest.version>
+
```
DarthSeppius: do you mind to add some session separator like `<-- Maven plugins version -->`? thx

yoti-sdk-parent/pom.xml
```diff
@@ -0,0 +1,419 @@
...
+      </dependency>
+      <dependency>
+        <groupId>javax.servlet</groupId>
+        <artifactId>servlet-api</artifactId>
+        <version>${javax.servlet.version}</version>
+      </dependency>
+  
+      <!-- Test dependencies -->
+      <dependency>
+        <groupId>junit</groupId>
```
DarthSeppius: should they have the scope `test` right? Or every time that they are declared on the child pom we explicitly add the scope there?

yoti-sdk-parent/pom.xml
```diff
@@ -0,0 +1,419 @@
...
+    <repository>
+      <id>ossrh</id>
+      <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
+    </repository>
+  </distributionManagement>
+  
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
+    <yoti.java.sdk.version>2.0.0-SNAPSHOT</yoti.java.sdk.version>
```
irotech: u could use ${project.version} here

yoti-sdk-parent/pom.xml
bucky-boy: <scope> is set in the child-projects.  If I try to control that in the parent pom I may come to regret it

yoti-sdk-parent/pom.xml
bucky-boy: That's better, thank you

yoti-sdk-parent/pom.xml
bucky-boy: I've made some changes - could you look again please, see if that's better?

PR #60, opened 2018-06-13
SDK-385: Introduce generic Attribute<T>

bucky-boy:
These changes are summarised in ticket SDK-385


yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/HumanProfileAdapter.java
```diff
@@ -26,16 +23,14 @@
     private static final String ATTRIBUTE_DOB = "date_of_birth";
     private static final String ATTRIBUTE_AGE_OVER = "age_over:";
     private static final String ATTRIBUTE_AGE_UNDER = "age_under:";
-    private static final String ATTRIBUTE_GENDER = "gender";
+    public static final String ATTRIBUTE_GENDER = "gender";
```
DarthSeppius: do you mind to group the public and private fields together pls?

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/SecureYotiClient.java
```diff
@@ -159,48 +161,35 @@ private EncryptedData parseProfileContent(byte[] profileContent) throws ProfileE
...
         for (AttrProto.Attribute attribute : message.getAttributesList()) {
             try {
-                parsedAttributes.add(AttributeConverter.convertAttribute(attribute));
-            } catch (IOException e) {
-                LOG.info("Cannot decode value for attribute {}", attribute.getName());
-            } catch (ParseException e) {
-                LOG.info("Cannot parse value for attribute {}", attribute.getName());
+                parsedAttributes.add(attributeConverter.convertAttribute(attribute));
+            } catch (IOException | ParseException e) {
+                LOG.warn("Cannot parse value for attribute '{}'", attribute.getName());
```
DarthSeppius: I know that is like this before but I'd like to keep the exception in the logs to see what could be the cause, do you agree?

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/SimpleProfileTest.java
```diff
@@ -1,211 +1,132 @@
...
+    public void getAttributeTyped_shouldThrowExceptionForForMismatchingType() {
+        SimpleProfile profile = new SimpleProfile(asAttributeList(SOME_KEY, INTEGER_VALUE));
 
-        Integer result = profile.getAttribute(SOME_KEY, Integer.class);
-
-        assertEquals(INTEGER_VALUE, result);
+        try {
+            profile.getAttribute(SOME_KEY, String.class);
+        } catch (ClassCastException e) {
+            assertTrue(e.getMessage().contains("java.lang.String"));
```
DarthSeppius: Is it good to rely on the message for the test assertion? It's not better to catch the exception and check if it is of the type that we wanted?

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/HumanProfileAdapter.java
bucky-boy: Maybe a better fix is to pull all the Attribute names out into one class, and make them all public...?

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/SecureYotiClient.java
bucky-boy: I've replied in the SDK-385 ticket rather than here in the public repo...

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/SimpleProfileTest.java
bucky-boy: It does catch ClassCastException, which is what we are looking for.  
Checking the message was just to ensure I'd used the correct the method on Class (.getCanonicalName())

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/HumanProfileAdapter.java
DarthSeppius: This is also a good option

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/SimpleProfileTest.java
qzagarese: then you can use `String.class.getName()`

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/SimpleProfileTest.java
```diff
@@ -1,211 +1,132 @@
...
+            assertTrue(e.getMessage().contains("java.lang.Integer"));
```
qzagarese: same here

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/SimpleProfileTest.java
```diff
@@ -1,211 +1,132 @@
...
-    public void findAttributeStartingWith_shouldReturnNullForMismatchingType() {
-        Profile profile = new SimpleProfile(asList(STARTS_WITH, INTEGER_VALUE));
+    public void findAttributeStartingWith_shouldThrowExceptionForMismatchingType() {
+        SimpleProfile profile = new SimpleProfile(asAttributeList(STARTS_WITH, INTEGER_VALUE));
 
-        String result = profile.findAttributeStartingWith(STARTS_WITH, String.class);
+        try {
+            profile.findAttributeStartingWith(STARTS_WITH, String.class);
+        } catch (ClassCastException e) {
+            assertTrue(e.getMessage().contains("java.lang.String"));
```
qzagarese: same here

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/SimpleProfileTest.java
```diff
@@ -1,211 +1,132 @@
...
+            assertTrue(e.getMessage().contains("java.lang.Integer"));
```
qzagarese: same here

PR #61, opened 2018-06-15
SDK-389: Break up SecureYotiClient

bucky-boy:
Some big changes, but there should be no change to the API or the behaviour of the SDK to anyone using it.

More detail in the ticket.

### DIFF
--- a/yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/ActivityDetailsFactory.java
+++ b/yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/ActivityDetailsFactory.java
@@ -18,9 +18,9 @@ import com.yoti.api.client.ActivityDetails;
 import com.yoti.api.client.Profile;
 import com.yoti.api.client.ProfileException;
 import com.yoti.api.client.spi.remote.call.Receipt;
-import com.yoti.api.client.spi.remote.util.Decryption;
+import com.yoti.api.client.spi.remote.util.DecryptionHelper;
 
-public class ActivityDetailsFactory {
+class ActivityDetailsFactory {
 
     private static final String RFC3339_PATTERN = "yyyy-MM-dd'T'HH:mm:ss'Z'";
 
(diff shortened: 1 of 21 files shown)


yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/ActivityDetailsFactory.java
```diff
@@ -0,0 +1,66 @@
...
+
+import javax.crypto.spec.SecretKeySpec;
+
+import com.yoti.api.client.ActivityDetails;
+import com.yoti.api.client.Profile;
+import com.yoti.api.client.ProfileException;
+import com.yoti.api.client.spi.remote.call.Receipt;
+import com.yoti.api.client.spi.remote.util.Decryption;
+
+public class ActivityDetailsFactory {
```
qzagarese: Why is this Factory public? 
It really shouldn't be used from client code.

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/util/Decryption.java
```diff
@@ -0,0 +1,29 @@
...
+import static com.yoti.api.client.spi.remote.call.YotiConstants.BOUNCY_CASTLE_PROVIDER;
+
+import java.security.GeneralSecurityException;
+import java.security.PrivateKey;
+
+import javax.crypto.Cipher;
+
+import com.yoti.api.client.ProfileException;
+
+public class Decryption {
```
qzagarese: can we come up with a different kind of naming? `CryptoHelper` or something along those lines?

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/ActivityDetailsFactory.java
bucky-boy: Fixed that one, and a few others where public access is not required.

This does force us to keep components in the same package, which is not ideal, but if we want to restrict access I don't see another way of doing it.

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/util/Decryption.java
bucky-boy: It's now named DecryptionHelper.

PR #62, opened 2018-06-19
SDK-394: Stop using Enums for attribute values

bucky-boy:
I've written quite a bit in the ticket.  In short...

We will not use Enums for attribute values.  If we start to return a new value for a given attribute, say a new _Gender_ type or a new _DocumentType_ then parsing that new value will cause an error.  Users of the SDK will be forced to upgrade in order to access the new value, and that's what we're trying to avoid.

### DIFF
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/HumanProfile.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/HumanProfile.java
@@ -7,34 +7,39 @@ import java.util.Map;
  */
 public interface HumanProfile extends Profile {
 
+    String GENDER_MALE = "MALE";
+    String GENDER_FEMALE = "FEMALE";
+    String GENDER_TRANSGENDER = "TRANSGENDER";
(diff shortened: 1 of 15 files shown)


yoti-sdk-api/src/main/java/com/yoti/api/client/HumanProfile.java
```diff
@@ -7,34 +7,39 @@
  */
 public interface HumanProfile extends Profile {
 
+    String GENDER_MALE = "MALE";
```
qzagarese: I think you were planning to have these as constants.
If so, please update accordingly.

bucky-boy: That is a constant.  _public final_ is redundant, because its being defined on an interface (public + final by default)

qzagarese: sorry, hadn't seen it was an interface, my bad :-(

PR #63, opened 2018-06-22
SDK-397: Start returning Anchor objects for Attributes

bucky-boy:
Details in the ticket....
https://lampkicking.atlassian.net/browse/SDK-397


yoti-sdk-api/src/main/java/com/yoti/api/client/DocumentDetails.java
```diff
@@ -6,12 +6,17 @@
  */
 public interface DocumentDetails {
 
+    String DOCUMENT_TYPE_PASSPORT = "PASSPORT";
```
qzagarese: they are supposedly constants

yoti-sdk-api/src/main/java/com/yoti/api/client/HumanProfile.java
```diff
@@ -7,34 +7,39 @@
  */
 public interface HumanProfile extends Profile {
 
+    String GENDER_MALE = "MALE";
```
qzagarese: these should be constants

yoti-sdk-api/src/main/java/com/yoti/api/client/Time.java
```diff
@@ -0,0 +1,13 @@
...
+public interface Time {
+
+    int getHour();
+
+    int getMinute();
+
+    int getSecond();
+
+    int getMillisecond();
+
```
qzagarese: The signed timestamp resolution is microseconds, so you might want to add it.

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/TimeValue.java
```diff
@@ -0,0 +1,55 @@
...
+    private final int minute;
+    private final int second;
+    private final int millisecond;
+
+    private TimeValue(int hour, int minute, int second, int millisecond) {
+        this.hour = hour;
+        this.minute = minute;
+        this.second = second;
+        this.millisecond = millisecond;
+    }
```
qzagarese: same here, you might want to introduce microseconds.

yoti-sdk-api/src/main/java/com/yoti/api/client/DocumentDetails.java
bucky-boy: That is a constant - its declared on an interface (public + final are not required)

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/TimeValue.java
bucky-boy: Ok, have switched from milliseconds to microseconds

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/AnchorConverter.java
```diff
@@ -0,0 +1,109 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class AnchorConverter {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AnchorConverter.class);
+
+    Anchor convert(AttrProto.Anchor anchorProto) throws CertificateException, IOException {
+        List<X509Certificate> certificates = convertCertificates(anchorProto);
+        CertificateType certificateType = determineAnchorType(certificates);
```
DarthSeppius: The name `CertificateType` is a little bit misleading, also we use a method that is called `determineAnchorType` to get ti back from a List of X509Certs. What do you mean for CertificateType here? the particular subset of Anchor type and value that are needed for creating the Anchor?

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/SignedTimestampValue.java
```diff
@@ -0,0 +1,24 @@
+package com.yoti.api.client.spi.remote;
+
+import com.yoti.api.client.DateTime;
+import com.yoti.api.client.SignedTimestamp;
+
+public class SignedTimestampValue implements SignedTimestamp {
```
DarthSeppius: `DefaultSignedTimestamp` or `SimpleSignedTimestamp`? Do we have other implementation of this interface? SignedTimestampValue seems a strange name.

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/AnchorConverterTest.java
```diff
@@ -0,0 +1,164 @@
...
+    public void shouldConvertPassportAnchor() throws Exception {
+        AttrProto.Anchor anchorProto = AttrProto.Anchor.parseFrom(Base64.getDecoder().decode(TestAnchors.PP_ANCHOR));
+
+        Anchor result = testObj.convert(anchorProto);
+
+        assertEquals("SOURCE", result.getType());
+        assertEquals("OCR", result.getSubType());
+        assertEquals(PASSPORT_SOURCE_TYPE, result.getValue());
+        verifyCertificates(result.getOriginCertificates(), PASSPORT_ISSUER, AnchorType.SOURCE.extensionOid);
+        verifyTimestamp(result.getSignedTimestamp(), "2018-Apr-12 13:14:32.835537");
```
DarthSeppius: avoid magic string or values, it's better to have is in a local or constant strings

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/SignedTimestampValue.java
bucky-boy: It follows the pattern in place for _DateValue implements Date_.

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/Receipt.java
```diff
@@ -95,12 +95,12 @@ public Receipt build() {
         }
     }
 
-    public static enum Outcome {
+    public enum Outcome {
```
irotech: didn't we drop all the enums?

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/AnchorConverterTest.java
bucky-boy: Ok.  I've moved these String Constants out into the TestAnchors file, and grouped them with the Anchor they are from.

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/Receipt.java
bucky-boy: Correct.  We did.  Well spotted.
The idea is that we will not use Enums for attribute values - we will not return Enum's to SDK clients.
This one here is used internally.  Values from the Receipt will be mapped to some implementation of _Profile_, which will not use Enums.  I think this should be ok.  There are similar uses of Enums in the xxxProto classes, which we have no control over.

yoti-sdk-api/src/main/java/com/yoti/api/client/DocumentDetails.java
qzagarese: sorry, hadn't realised it was an interface. My bad :-(

yoti-sdk-api/src/main/java/com/yoti/api/client/HumanProfile.java
qzagarese: same here, sorry

PR #68, opened 2018-06-27
SDK-390: Update documentation for v2

bucky-boy:
Mostly focussed on the readme, but did some api javadocs too

https://lampkicking.atlassian.net/browse/SDK-390

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,9 @@ Description of utilising Spring Boot
 1) [Running the Example](#running-the-example) -
 Running the Spring example project
 
+1) [Breaking changes and enhancements made in v2.0.0](#breaking-changes-and-enhancements-made-in-v2.0.0) -
+Things you need to know when migrating from 1.x to 2.x
+
 1) [Spring Security Integration](#spring-security-integration) -
 Integrating Yoti Authentication with Spring Boot.
 
@@ -58,7 +61,7 @@ To integrate your application with Yoti, your back-end must expose a GET endpoin
 The endpoint can be configured in Yoti Dashboard when you create/update your application.
 
 The image below shows how your application back-end and Yoti integrate in the context of a Login flow.
-Yoti SDK carries out steps 6 through 9 for you, including profile decryption and communication with backend services.
+Use the Yoti SDK to complete steps 6 through 9 for you, including profile decryption and communication with backend services.
 
 ![alt text](login_flow.png "Login flow")
 
@@ -66,20 +69,20 @@ Yoti also allows you to enable user details verification from your mobile app by
 
 ## Requirements
 
-* Java 1.6 or higher
+* Java 1.7 or higher
 * SLF4J
 
 ## Building From Source
 
-Building from source is generally not necessary for third parties since artifacts are published in Maven Central. However, if you want to build from source you can do so using the [Maven Wrapper](https://github.com/takari/maven-wrapper) that is bundled with this distribution. For those familiar with Gradle this is much like the Gradle Wrapper and ensures that the correct version of Maven is being used.
+Building from source is not necessary since artifacts are published in Maven Central. However, if you want to build from source use the [Maven Wrapper](https://github.com/takari/maven-wrapper) bundled with this distribution. For those familiar with Gradle this is much like the Gradle Wrapper and ensures that the correct version of Maven is being used.
 
 From the top level directory:
 
 ```bash
-./mvnw clean install
+./mvnw clean install -Dgpg.skip=true
 ```
 
-Notable flags that you may wish to use to skip certain static analysis/code quality tools are listed below. This is only recommended if you find that these tools are taking too long during development or are flagging false positives that you are yet to exclude. **They should not be ignored when building a candidate for a release unless you are sure that the issues being raised are not a cause for concern.**
+Notable flags that you may wish to use to skip certain static analysis/code quality tools are listed below.
 
 * `-Dfindbugs.skip=true`: skips findbugs and the findbugs security extension.
 * `-Ddependency-check.skip=true`: skips the OWASP dependency scanner.
@@ -92,26 +95,26 @@ Notable flags that you may wish to use to skip certain static analysis/code qual
 
 ## Enabling the SDK
 
-To import the Yoti SDK inside your project, you can use your favourite dependency management system.
+To include the Yoti SDK in your project, you can use your favourite dependency management system.
 If you are using Maven, you need to add the following dependency:
 
 ```xml
 <dependency>
     <groupId>com.yoti</groupId>
     <artifactId>yoti-sdk-impl</artifactId>
-    <version>1.5.0</version>
+    <version>2.0.0</version>
 </dependency>
 ```
 
 If you are using Gradle, here is the dependency to add:
 
-`compile group: 'com.yoti', name: 'yoti-sdk-impl', version: '1.5.0'`
+`compile group: 'com.yoti', name: 'yoti-sdk-impl', version: '2.0.0'`
 
 You will find all classes packaged under `com.yoti.api`
 
 ## Client Initialisation
 
-The YotiClient is the SDK entry point. To initialise it you need include the following snippet inside your endpoint initialisation section:
+The entry point of the SDK is `com.yoti.api.client.YotiClient`.  To initialise it you need to include the following snippet:
 
 ```java
 import com.yoti.api.client.YotiClient;
@@ -120,7 +123,8 @@ import static com.yoti.api.client.FileKeyPairSource.fromFile;
 
 YotiClient client = YotiClientBuilder.newInstance()
     .forApplication(<YOUR_CLIENT_SDK_ID>)
-    .withKeyPair(fromFile(new java.io.File("<PATH/TO/YOUR/APPLICATION/KEY_PAIR.pem>"))).build();
+    .withKeyPair(fromFile(new java.io.File("<PATH/TO/YOUR/APPLICATION/KEY_PAIR.pem>")))
+    .build();
 ```
 
 Where:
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -130,7 +134,19 @@ Where:
 
 ## Profile Retrieval
 
-When your application receives a token via the exposed endpoint (it will be assigned to a query string parameter named `token`), you can easily retrieve the user profile by adding the following to your endpoint handler:
+### Callback URl
+
+To allow sharing a Users profile with your application, you must expose a GET endpoint over https that Yoti will use to push a one-time token to your back end.
```
echarrod: We've gone with `one time use token` elsewhere, so would be good to have that here too instead of `one-time token` for consistency. Also can we capitalise `HTTPS`?

README.md
```diff
@@ -130,7 +134,19 @@ Where:
...
+
+You can use any address, path or port in your URL, but it must be https and it must allow a query string parameter named `token` - that's the value you wil need for retrieving the shared profile.
+
+You configure the callback URL for your application on the Integration page of the Yoti Dashboard.
+
+**Important: these tokens only allow a single use. You cannot retrieve a profile using the same token more than once.**
+
+### ActivityDetails
+
+When your application receives a one-time token from Yoti, you retrieve details of the share with the following:
```
echarrod: Ditto

README.md
```diff
@@ -193,14 +209,31 @@ try {
...
+
+### Anchors, Sources and Verifiers
 
-The `com.yoti.api.client.HumanProfile` class provides a set of methods to retrieve different user attributes. Whether the attributes are present or not depends on the settings you have applied to your app on Yoti Dashboard.
+`com.yoti.api.client.Anchor` represents how a given `Attribute<T>` has been _sourced_ or _verified_.  These values are created and signed whenever a Profile Attribute is created, or verified with an external party.
+
+For example, an attribute value that was _sourced_ from a Passport might have the following values:
+| attribute.getType | SOURCE |
+| attribute.getValue | PASSPORT |
+| attribute.getTimestamp | 2017-10-31, 19:45:59.123789 |
```
echarrod: Since type, value & timestamp are associated with an anchor and not an attribute, shouldn't this be something like:

For example, an attribute value that was _sourced_ from a Passport might have the following **anchor properties**:

and then:
attribute.**anchor**.getType,
attribute.**anchor**.getValue,
attribute.**anchor**.getTimestamp?

README.md
```diff
@@ -295,6 +330,27 @@ For more information and to see an example of this in use take a look at the [Sp
 
 Instructions on how to run the Spring example project can be found in the [yoti-sdk-spring-boot-example](/yoti-sdk-spring-boot-example) folder.
 
+## Breaking changes and enhancements made in v2.0.0
+
+As well as exposing significant new functionality, we've taken this opportunity to clean up and simply our public API.
```
echarrod: *simplify

README.md
```diff
@@ -295,6 +330,27 @@ For more information and to see an example of this in use take a look at the [Sp
 
 Instructions on how to run the Spring example project can be found in the [yoti-sdk-spring-boot-example](/yoti-sdk-spring-boot-example) folder.
 
+## Breaking changes and enhancements made in v2.0.0
+
```
echarrod: Since the README is already quite large, do you think it would be better to put these changes in a CHANGELOG.md (like we have in [.NET](https://github.com/getyoti/yoti-dotnet-sdk/blob/master/CHANGELOG.md), Python, Ruby & Node).

We can place a link to the changelog in the README still. This would mean the README wouldn't have to grow with the changes we make for each new version.

yoti-sdk-api/src/main/java/com/yoti/api/client/Anchor.java
```diff
@@ -5,14 +5,40 @@
 
 public interface Anchor {
 
+    /**
+     * The type of Anchor - most likely either SOURCE or VERIFIER, but its possible that new Anchor types will be added in future
```
echarrod: *it's

yoti-sdk-api/src/main/java/com/yoti/api/client/Anchor.java
```diff
@@ -5,14 +5,40 @@
...
+     *
+     * @return String, naming the type of the anchor
+     */
     String getType();
 
+    /**
+     * Describes how the attribute value was obtained, for example this might be 'OCR' or 'NFC'
```
echarrod: This makes sense for sources, but the subtype for Verifiers isn't really how the attribute value was obtained is it? E.g. we can have a Verifier value of `YOTI_IDENTITY` with a subtype of `Callcredit`, but we are just verifying with Callcredit, not obtaining the value from them.

README.md
```diff
@@ -130,7 +134,19 @@ Where:
 
 ## Profile Retrieval
 
-When your application receives a token via the exposed endpoint (it will be assigned to a query string parameter named `token`), you can easily retrieve the user profile by adding the following to your endpoint handler:
+### Callback URl
+
+To allow sharing a Users profile with your application, you must expose a GET endpoint over https that Yoti will use to push a one-time token to your back end.
+
+You can use any address, path or port in your URL, but it must be https and it must allow a query string parameter named `token` - that's the value you wil need for retrieving the shared profile.
```
qzagarese: We need to rephrase this:
`You can use any public domain name or localhost, as well as any path or port, but the resulting url must be served over https. Your endpoint must consume a query string parameter named 'token' - that's the value you will need for retrieving the shared profile`

README.md
```diff
@@ -149,21 +165,21 @@ import com.yoti.api.client.ProfileException;
...
+    //use the profile to extract attributes
 } catch (ProfileException e) {
     LOG.info("Could not get profile", e);
-    // do something meaningful.
+    // do something meaningful
 }
 ```
 
-**Important: encrypted tokens are intended for single use only. You must not invoke the client using the same token more than once.**
+`com.yoti.api.client.ActivityDetails` gives you access to the `com.yoti.api.client.ApplicationProfile` of your app as well as the `com.yoti.api.client.HumanProfile` of the user that has shared with you.
```
qzagarese: you can remove the ApplicationProfile bit

README.md
```diff
@@ -177,7 +193,7 @@ try {
         String fullName = profile.GetFullName();
         String givenNames = profile.getGivenNames();
         String familyName = profile.getFamilyName();
-        String mobileNumber = profile.getPhoneNumber();
+        String phoneNumber = profile.getPhoneNumber();
```
qzagarese: This snippet should use the `Attribute` class now

README.md
```diff
@@ -193,14 +209,31 @@ try {
...
 
-The `com.yoti.api.client.HumanProfile` class provides a set of methods to retrieve different user attributes. Whether the attributes are present or not depends on the settings you have applied to your app on Yoti Dashboard.
+`com.yoti.api.client.Anchor` represents how a given `Attribute<T>` has been _sourced_ or _verified_.  These values are created and signed whenever a Profile Attribute is created, or verified with an external party.
+
+For example, an attribute value that was _sourced_ from a Passport might have the following values:
+| attribute.getType | SOURCE |
+| attribute.getValue | PASSPORT |
+| attribute.getTimestamp | 2017-10-31, 19:45:59.123789 |
+
+Similarly, an attribute _verified_ against the data held by an external party will have an `Anchor` of type _VERIFY_, naming the party that verified it.
```
qzagarese: `VERIFIER`

README.md
qzagarese: simplify

README.md
```diff
@@ -295,6 +330,27 @@ For more information and to see an example of this in use take a look at the [Sp
...
+As well as exposing significant new functionality, we've taken this opportunity to clean up and simply our public API.
+
+### Dropped support for Java 6
+Minimum supported Java version is now 7.
+
+### New interfaces - Attribute<T> and Anchor
+The old `com.yoti.api.client.Attribute` class has been replaced with `com.yoti.api.client.Attribute<T>`, which now exposes a `List<Anchor>` for the sources and verifiers of the attribute.
+The old `Set<String> getSources()` and `Set<String> getVerifiers()` methods have been replaced with `List<Anchor> getSources()` and `List<Anchor> getVerifiers()`.
+
+### Changes to HumanProfile and ApplicationProfile
```
qzagarese: You can remove ApplicationProfile

bucky-boy: Well, ApplicationProfile did change too....

README.md
bucky-boy: Ok.

README.md
bucky-boy: Correct.  That's wrong.  Fixed.

README.md
bucky-boy: Cheers, well spotted

README.md
bucky-boy: I'm not sure.  If we're moving towards using the Yoti website as the primary home for documentation, then these READMEs will probably only have stuff like change logs in them.

yoti-sdk-api/src/main/java/com/yoti/api/client/Anchor.java
bucky-boy: Thank you.

README.md
bucky-boy: Done.

README.md
bucky-boy: ok

README.md
bucky-boy: Good spot, thanks.

README.md
bucky-boy: ok

README.md
echarrod: We also have these changes in the release notes on Github (e.g. [.NET](https://github.com/getyoti/yoti-dotnet-sdk/releases)). I'm just thinking that it's polluting the README a bit if we do this every time, when we can represent this information in other places

bucky-boy: That does look nice.  We haven't been using those Releases in Java SDK - we should fix that.
But I'm still concerned about how many places we want people to look - we have 4x READMEs here, as well as the Yoti website.

README.md
qzagarese: :+1:

PR #69, opened 2018-06-28
SDK-387: Throw Exception for a malformed DocumentDetails attribute value

bucky-boy:
DocumentDetailsAttributeParsers shall throw an Exception if it receives a malformed DocumetDetails value that it cannot parse.
That Exception will be logged as a WARN by the AttributeListConverter.  The failed attribute is skipped, the Profile will have _null_ for the missing value.
Discussed further in the ticket.

### DIFF
--- a/yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/AttributeConverter.java
+++ b/yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/AttributeConverter.java
@@ -59,7 +59,7 @@ class AttributeConverter {
             case JSON:
                 return JSON_MAPPER.readValue(attribute.getValue().toString(DEFAULT_CHARSET), Map.class);
             default:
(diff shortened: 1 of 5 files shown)


yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/AttributeConverter.java
```diff
@@ -59,7 +59,7 @@ private Object convertValueFromProto(AttrProto.Attribute attribute) throws Parse
             case JSON:
                 return JSON_MAPPER.readValue(attribute.getValue().toString(DEFAULT_CHARSET), Map.class);
             default:
-                LOG.error("Unknown type {} for attribute {}", attribute.getContentType(), attribute.getName());
+                LOG.warn("Unknown type {} for attribute {}.  Attempting to parse it as a String", attribute.getContentType(), attribute.getName());
```
DarthSeppius: I think there is an extra space here.

PR #74, opened 2018-07-18
SDK-422: New profile for deploying to /nxs/repository/maven-snaphosts/

bucky-boy:
We shall have 2x profiles
- *nexus* for deploying to our internal nexus (use with -Dgpg.skip=true)
- *sonatype* for deploying signed JARs to public sonatype

### DIFF
--- a/yoti-sdk-parent/pom.xml
+++ b/yoti-sdk-parent/pom.xml
@@ -56,6 +56,10 @@
       <activation>
         <activeByDefault>false</activeByDefault>
       </activation>
+      <properties>
+        <gpg.skip>false</gpg.skip>
+        <dependency-check.skip>false</dependency-check.skip>
+      </properties>
       <distributionManagement>
         <snapshotRepository>
           <id>ossrh</id>
@@ -72,6 +76,10 @@
       <activation>
         <activeByDefault>false</activeByDefault>
       </activation>
+      <properties>
+        <gpg.skip>true</gpg.skip>
+        <dependency-check.skip>true</dependency-check.skip>
+      </properties>
       <distributionManagement>
         <snapshotRepository>
           <id>yoti-snapshots</id>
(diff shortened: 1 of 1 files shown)


pn-santos:
> use with -Dgpg.skip=true

You can set that property to true in the profile config with
```
<properties>
     <gpg.skip>true</gpg.skip>
<properties>
```

yoti-sdk-parent/pom.xml
```diff
@@ -50,17 +50,38 @@
...
-  <distributionManagement>
-    <snapshotRepository>
-      <id>ossrh</id>
-      <url>https://oss.sonatype.org/content/repositories/snapshots</url>
-    </snapshotRepository>
-    <repository>
-      <id>ossrh</id>
-      <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
-    </repository>
-  </distributionManagement>
```
pn-santos: I'd leave this here or activate the sonatype profile by default (alternatively the nexus could be the default, up to what you think best).

It's always better to have a default (usually you'd pick the one that get's used more often) and then have profiles for all other cases. Otherwise it forces users to always have to include a profile arg when running the mvn command.

yoti-sdk-parent/pom.xml
```diff
@@ -50,17 +50,38 @@
...
+          <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+        </snapshotRepository>
+        <repository>
+          <id>ossrh</id>
+          <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
+        </repository>
+      </distributionManagement>
+    </profile>
+    <profile>
+      <id>nexus</id>
```
pn-santos: you can set the `gpg.skip` property here to false

yoti-sdk-parent/pom.xml
bucky-boy: I'd rather not.  We only need the profiles if we specifically want to deploy to either sonatype or nexus, and doing either should not be frequent.  In this case I'm quite happy to be slightly awkward about it - they'd have to be following a release process in the wiki anyway.

PR #75, opened 2018-07-19
SDK-460: Fix display of sources and verifiers

bucky-boy:
This is an easy fix, but I don't know how this has happened.  The original story was SDK-391, and that even has a screen shot displaying sources + verifiers.

My best guess is this change somehow got lost in a merge/rebase before the old SNAPSHOT branch was merged to master.

### DIFF
--- a/yoti-sdk-spring-boot-example/src/main/resources/templates/fragments.html
+++ b/yoti-sdk-spring-boot-example/src/main/resources/templates/fragments.html
@@ -2,10 +2,12 @@
 <html xmlns:th="http://www.thymeleaf.org">
 <head/>
 <body>
-    <div class="" th:fragment="anchor">
-
-        Hello!
-
+<td th:fragment="anchorDisplay(anchors)">
+    <div th:each="anchor : ${anchors}" style="border: 1px solid grey; display:inline-block; margin-bottom=0" >
+        <p style="line-height:0"><i>Type:</i> <span th:text="${anchor.type}"/></p>
+        <p style="line-height:0"><i>Value:</i> <span th:text="${anchor.value}"/></p>
+        <p style="line-height:0"><i>SubType:</i> <span th:text="${anchor.subType}"/></p>
+        <p style="line-height:0"><i>Timestamp:</i> <span th:text="${anchor.signedTimestamp.timestamp.date}"/>, <span th:text="${anchor.signedTimestamp.timestamp.time}"/></p>
     </div>
 </body>
 </html>


yoti-sdk-spring-boot-example/src/main/resources/templates/fragments.html
```diff
@@ -2,10 +2,12 @@
...
-    <div class="" th:fragment="anchor">
-
-        Hello!
-
+<td th:fragment="anchorDisplay(anchors)">
+    <div th:each="anchor : ${anchors}" style="border: 1px solid grey; display:inline-block; margin-bottom=0" >
+        <p style="line-height:0"><i>Type:</i> <span th:text="${anchor.type}"/></p>
+        <p style="line-height:0"><i>Value:</i> <span th:text="${anchor.value}"/></p>
+        <p style="line-height:0"><i>SubType:</i> <span th:text="${anchor.subType}"/></p>
+        <p style="line-height:0"><i>Timestamp:</i> <span th:text="${anchor.signedTimestamp.timestamp.date}"/>, <span th:text="${anchor.signedTimestamp.timestamp.time}"/></p>
```
echarrod: I don't believe we have timestamp in the other SDK profile pages. It is a nice addition to have, I'm just thinking that the profile page is quite crowded already.

bucky-boy: Well, we return it.  So let's go with it for now.  Whether we keep it in the new demo design is another matter.  This is just an interim change.

yoti-sdk-spring-boot-example/src/main/resources/templates/fragments.html
```diff
@@ -2,10 +2,12 @@
 <html xmlns:th="http://www.thymeleaf.org">
 <head/>
 <body>
-    <div class="" th:fragment="anchor">
-
-        Hello!
-
+<td th:fragment="anchorDisplay(anchors)">
+    <div th:each="anchor : ${anchors}" style="border: 1px solid grey; display:inline-block; margin-bottom=0" >
```
sidibos: Is it better to move these styles inside a file `style.css` and include it, rather than including them inline?

bucky-boy: Yes, that would certainly be better, but since this is just an interim measure, let's not got that far.  

We expect to delete this very soon, as soon as we get the new demo page.

PR #79, opened 2018-08-30
NA: Fix table display in README.md

bucky-boy:


README.md
```diff
@@ -224,9 +224,13 @@ Attributes are returned as an instance of `com.yoti.api.client.Attribute<T>`.  S
...
-| anchor.getType | SOURCE |
-| anchor.getValue | PASSPORT |
-| anchor.getTimestamp | 2017-10-31, 19:45:59.123789 |
+
+`Anchor` property | Example value
+-----|------
+anchor.getType | SOURCE
+anchor.getValue | PASSPORT
+anchor.getSubType | OCR
+anchor.getSignedTimestamp | 2017-10-31, 19:45:59.123789
```
sidibos: Should this be anchor.getSignedTimeStamp.getTimestamp?

bucky-boy: Well spotted, thank you.  Yes it should.  I'll amend the commit.

bucky-boy: Actually, I'll change it to use field names and drop the method references.  It looks neater that way, and it's trivial for the reader to work out the getters they should use.

PR #80, opened 2018-09-03
SDK-538: Add support for Anchors etc to the sandbox module

bucky-boy:
Mostly low-level, builder stuff.

Note that the change in the _README_ is already merged into _master_.  It's here to bring the _DEVELOPMENT_ branch into line.  It will drop out when we pull + rebase _DEVELOPMENT_ onto _master_.

### DIFF
--- a/yoti-sdk-sandbox/src/main/java/com/yoti/api/client/sandbox/profile/request/attribute/SandboxAnchor.java
+++ b/yoti-sdk-sandbox/src/main/java/com/yoti/api/client/sandbox/profile/request/attribute/SandboxAnchor.java
@@ -1,0 +1,103 @@
+package com.yoti.api.client.sandbox.profile.request.attribute;
+
+import java.util.Calendar;
+import java.util.GregorianCalendar;
+import java.util.TimeZone;
+
+import com.yoti.api.client.Date;
+import com.yoti.api.client.DateTime;
+import com.yoti.api.client.Time;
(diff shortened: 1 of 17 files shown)


yoti-sdk-sandbox/src/main/java/com/yoti/api/client/sandbox/profile/request/attribute/SandboxAnchor.java
```diff
@@ -0,0 +1,103 @@
...
+    public String getSubType() {
+        return subType;
+    }
+
+    @JsonProperty("timestamp")
+    public long getTimestamp() {
+        return timestamp;
+    }
+
+    public static class SandboxAnchorBuilder {
```
DarthSeppius: if we are going to expose the builder only thought 
```  
  public static SandboxAnchor.SandboxAnchorBuilder builder() {
        return new SandboxAnchor.SandboxAnchorBuilder();
  }
```
I think that this should be `private`

bucky-boy: But if the builder class is _private_, I can't use it outside of _SandboxAnchor_.

DarthSeppius: Aaaaand you are right! I'll hide the comment xD

PR #82, opened 2018-09-12
SDK-573: Introduce List<AgeVerification>

bucky-boy:
What do we think of this approach....?

We've decided we want to add support for multiple age-derived-attributes (age_over + age_under) now.  So we have a new _AgeVerification_ interface, and the _HumanProfile_ will now return a list of those wrapping all the age-derived-attributes that it can find.  There are also two helper methods to sweeten it, _findAgeOverVerification(int age)_ and _findAgeUnderVerification(int age)_.

The real meat of this change can be found in _AgeVerification_, _HumanProfile_ and _HumanProfileAdapter_.

Introducing _AgeVerification_ to the api package prompts us to rename the class we already have in the sandbox: _AgeVerification_ > _SandboxAgeVerification_.  This is not a breaking change, since none of the sandbox code has been released yet.  But it will require a change in the _connect-api-sandbox_ Integration tests - coming very soon.

When updating the README and the JAVADOC I notices a few other problems hanging over from the Java SDK v2 change.  I've taken the opportunity to fix them now.

Finally, although we don't have any spec for it, I think it would be prudent to prove that our implementation of DocumentDetails does break if we start receiving a 6th parameter.  So there's a unit test for that too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,13 +190,18 @@ try {
         String rememberMeId = activityDetails.getRememberMeId();
         String base64Selfie = activityDetails.getBase64Selfie();
         Attribute<Image> selfie = profile.getSelfie();
-        Image selfieValue = selfie.getValue();
+        if (selfie != null) {
+            Image selfieValue = selfie.getValue();
+        }
         Attribute<String> fullName = profile.getFullName();
         Attribute<String> givenNames = profile.getGivenNames();
         Attribute<String> familyName = profile.getFamilyName();
         Attribute<String> phoneNumber = profile.getPhoneNumber();
         Attribute<String> emailAddress = profile.getEmailAddress();
-        Boolean isAgeVerified = profile.isAgeVerified();
+        AgeVerification over18Verification = profile.findAgeOverVerification(18);
+        if (over18Verification != null) {
+            boolean isAgedOver18 = over18Verification.getResult();
+        }
         Attribute<Date> dateOfBirth = profile.getDateOfBirth();
         Attribute<String> gender = profile.getGender();
         Attribute<String> postalAddress = profile.getPostalAddress();
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/AgeVerification.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/AgeVerification.java
@@ -1,0 +1,36 @@
+package com.yoti.api.client;
+
+/**
+ * Wraps an 'Age Verify/Condition' attribute to provide behaviour specific to verifying someone's age
+ */
+public interface AgeVerification {
+
+    /**
+     * The type of age check performed, as specified on dashboard.  Currently this might be 'age_over' or 'age_under'
+     *
+     * @return The type of age check performed
+     */
+    String getCheckPerformed();
+
+    /**
+     * The age that was that checked, as specified on dashboard.
+     *
+     * @return The age that was that checked
+     */
+    int getAgeVerified();
+
+    /**
+     * Whether or not the profile passed the age check
+     *
+     * @return The result of the age check.
+     */
+    boolean getResult();
+
+    /**
+     * The wrapped profile attribute.  Use this if you need access to the underlying List of {@link Anchor}s
+     *
+     * @return The wrapped profile attribute
+     */
+    Attribute<String> getAttribute();
+
+}
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/AgeVerification.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/AgeVerification.java
@@ -6,7 +6,7 @@ package com.yoti.api.client;
 public interface AgeVerification {
 
     /**
-     * The type of age check performed, as specified on dashboard.  Currently this might be 'age_over' or 'age_under'
+     * The type of age check performed, as specified on dashboard. Currently this might be 'age_over' or 'age_under'
      *
      * @return The type of age check performed
      */
@@ -27,7 +27,7 @@ public interface AgeVerification {
     boolean getResult();
 
     /**
-     * The wrapped profile attribute.  Use this if you need access to the underlying List of {@link Anchor}s
+     * The wrapped profile attribute. Use this if you need access to the underlying List of {@link Anchor}s
      *
      * @return The wrapped profile attribute
      */
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/Anchor.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/Anchor.java
@@ -4,7 +4,7 @@ import java.security.cert.X509Certificate;
 import java.util.List;
 
 /**
- * Anchors represent where, when and how a profile {@link Attribute} value was acquired or verified.  For example, Attributes sourced from a backing
+ * Anchors represent where, when and how, a profile {@link Attribute} value was acquired or verified. For example, Attributes sourced from a backing
  * document such as a Passport will have a signed, timestamped anchor identifying that source
  */
 public interface Anchor {
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/Anchor.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/Anchor.java
@@ -3,6 +3,10 @@ package com.yoti.api.client;
 import java.security.cert.X509Certificate;
 import java.util.List;
 
+/**
+ * Anchors represent where, when and how a profile {@link Attribute} value was acquired or verified.  For example, Attributes sourced from a backing
+ * document such as a Passport will have a signed, timestamped anchor identifying that source
+ */
 public interface Anchor {
 
     /**
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/HumanProfile.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/HumanProfile.java
@@ -64,7 +64,7 @@ public interface HumanProfile extends Profile {
     AgeVerification findAgeUnderVerification(int age);
 
     /**
-     * @deprecated  Did the user pass the age verification check?  From SDK 2.1 onwards use getAgeVerifications(), findAgeOverVerification(int age)
+     * @deprecated Did the user pass the age verification check?  From SDK 2.1 onwards use getAgeVerifications(), findAgeOverVerification(int age)
      * or findAgeUnderVerification(int age)
      *
      * @return <code>TRUE</code> if they passed, <code>FALSE</code> if they failed, <code>null</code> if there was no check
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/HumanProfile.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/HumanProfile.java
@@ -1,9 +1,10 @@
 package com.yoti.api.client;
 
+import java.util.List;
 import java.util.Map;
 
 /**
- * Profile of an human user with convenience methods to access well-known attributes.
+ * Profile of a human user with convenience methods to access well-known attributes.
  */
 public interface HumanProfile extends Profile {
 
@@ -41,10 +42,34 @@ public interface HumanProfile extends Profile {
     Attribute<Date> getDateOfBirth();
 
     /**
-     * Did the user pass the age verification check?
+     * Finds all the 'Age Over' and 'Age Under' derived attributes returned with the profile, and returns them wrapped in {@link AgeVerification}
+     * objects
+     *
+     * @return A list of all the {@link AgeVerification} objects returned for this profile
+     */
+    List<AgeVerification> getAgeVerifications();
+
+    /**
+     * Searches for an {@link AgeVerification} corresponding to an 'Age Over' check for the given age
+     *
+     * @return age_over {@link AgeVerification} for the given age, or <code>null</code> if no match was found
+     */
+    AgeVerification findAgeOverVerification(int age);
+
+    /**
+     * Searches for an {@link AgeVerification} corresponding to an 'Age Under' check for the given age
+     *
+     * @return age_under {@link AgeVerification} for the given age, or <code>null</code> if no match was found
+     */
+    AgeVerification findAgeUnderVerification(int age);
+
+    /**
+     * @deprecated  Did the user pass the age verification check?  From SDK 2.1 onwards use getAgeVerifications(), findAgeOverVerification(int age)
+     * or findAgeUnderVerification(int age)
      *
      * @return <code>TRUE</code> if they passed, <code>FALSE</code> if they failed, <code>null</code> if there was no check
      */
+    @Deprecated
     Boolean isAgeVerified();
 
     /**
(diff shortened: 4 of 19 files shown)


yoti-sdk-api/src/main/java/com/yoti/api/client/Anchor.java
```diff
@@ -3,6 +3,10 @@
 import java.security.cert.X509Certificate;
 import java.util.List;
 
+/**
+ * Anchors represent where, when and how a profile {@link Attribute} value was acquired or verified.  For example, Attributes sourced from a backing
```
DarthSeppius: do you mind to add a `,` here `Anchors represent where, when, and how....`? Also, there is one more space after the first sentence.

yoti-sdk-api/src/main/java/com/yoti/api/client/AgeVerification.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * Whether or not the profile passed the age check
+     *
+     * @return The result of the age check.
+     */
+    boolean getResult();
+
+    /**
+     * The wrapped profile attribute.  Use this if you need access to the underlying List of {@link Anchor}s
```
DarthSeppius: Do you mind to remove the extra space after the first sentence? thx

yoti-sdk-api/src/main/java/com/yoti/api/client/HumanProfile.java
```diff
@@ -41,10 +42,34 @@
...
+
+    /**
+     * Searches for an {@link AgeVerification} corresponding to an 'Age Under' check for the given age
+     *
+     * @return age_under {@link AgeVerification} for the given age, or <code>null</code> if no match was found
+     */
+    AgeVerification findAgeUnderVerification(int age);
+
+    /**
+     * @deprecated  Did the user pass the age verification check?  From SDK 2.1 onwards use getAgeVerifications(), findAgeOverVerification(int age)
```
DarthSeppius: same here.

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/HumanProfileAdapterTest.java
```diff
@@ -59,4 +81,80 @@ public void isAgeVerified_shouldPreferAgeOverWhenBothAttributesArePresent() {
...
+        testObj.getAgeVerifications();
+        List<AgeVerification> result = testObj.getAgeVerifications();
+
+        assertThat(result, hasSize(4));
+        assertThat(new HashSet<>(result), hasSize(4));
+        assertThat(result.get(0).getAttribute(), isOneOf(ageUnder18Attribute, ageUnder21Attribute, ageOver18Attribute, ageOver21Attribute));
+        assertThat(result.get(1).getAttribute(), isOneOf(ageUnder18Attribute, ageUnder21Attribute, ageOver18Attribute, ageOver21Attribute));
+        assertThat(result.get(2).getAttribute(), isOneOf(ageUnder18Attribute, ageUnder21Attribute, ageOver18Attribute, ageOver21Attribute));
+        assertThat(result.get(3).getAttribute(), isOneOf(ageUnder18Attribute, ageUnder21Attribute, ageOver18Attribute, ageOver21Attribute));
+        verify(profileMock, times(1)).findAttributesStartingWith(HumanProfileAttributes.AGE_UNDER, String.class);
```
DarthSeppius: you can just do verify without times, as 1 is the default value.

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/HumanProfileAdapterTest.java
```diff
@@ -59,4 +81,80 @@ public void isAgeVerified_shouldPreferAgeOverWhenBothAttributesArePresent() {
...
+    @Test
+    public void findAgeOverVerification_shouldOnlySearchWrappedProfileOnce() {
+        when(profileMock.findAttributesStartingWith(HumanProfileAttributes.AGE_UNDER, String.class)).thenReturn(ageUnderAttributes);
+        when(profileMock.findAttributesStartingWith(HumanProfileAttributes.AGE_OVER, String.class)).thenReturn(ageOverAttributes);
+
+        testObj.findAgeOverVerification(18);
+        AgeVerification result = testObj.findAgeOverVerification(18);
+
+        assertSame(ageOver18Attribute, result.getAttribute());
+        verify(profileMock, times(1)).findAttributesStartingWith(HumanProfileAttributes.AGE_UNDER, String.class);
```
DarthSeppius: same here

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/HumanProfileAdapterTest.java
bucky-boy: True.  But writing it this way makes it explicit.

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/HumanProfileAdapterTest.java
bucky-boy: same response - it's particularly relevant, so I'd rather leave it

README.md
```diff
@@ -190,13 +190,18 @@ try {
         String rememberMeId = activityDetails.getRememberMeId();
         String base64Selfie = activityDetails.getBase64Selfie();
         Attribute<Image> selfie = profile.getSelfie();
-        Image selfieValue = selfie.getValue();
+        if (selfie != null) {
+            Image selfieValue = selfie.getValue();
+        }
         Attribute<String> fullName = profile.getFullName();
         Attribute<String> givenNames = profile.getGivenNames();
         Attribute<String> familyName = profile.getFamilyName();
```
echarrod: In the other SDKs we show how to return the values of each attribute here, rather than the Attribute object.

I like returning the attributes here in the way that you've done, as then there's not the potential to error when accessing the value if the attribute isn't present. Should we change the other SDKs to do it in this manner?

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/HumanProfileAdapter.java
```diff
@@ -50,70 +60,110 @@ public Attribute getAttribute(String name) {
...
+                return ageVerification;
+            }
+        }
+        return null;
+    }
+
+    private void findAllAgeVerifications() {
+        if (ageVerifications == null) {
+            List<AgeVerification> verifications = new ArrayList<>();
+            for (Attribute<String> ageOver : wrapped.findAttributesStartingWith(HumanProfileAttributes.AGE_OVER, String.class)) {
```
sidibos: I think it's better to have a rule for each age derivation and have a processor that will apply those rules accordingly.
So we can easily scale the age verification process should there be a new derivation.
An example can be seen here:
[Age Processor](https://github.com/getyoti/yoti-php-sdk/blob/master/src/Yoti/Util/Age/Processor.php)
[Age Under|Over Processor](https://github.com/getyoti/yoti-php-sdk/blob/master/src/Yoti/Util/Age/AgeUnderOverProcessor.php)
If we have a new derivation like age range, we could have a new Processor for it, which can be added to the process list [here](https://github.com/getyoti/yoti-php-sdk/blob/master/src/Yoti/Util/Age/Processor.php#L62)

yoti-sdk-api/src/main/java/com/yoti/api/client/Profile.java
```diff
@@ -1,44 +1,55 @@
...
+     * Returns the {@link Attribute} object for the key
      *
      * @param name
      * @return the attribute object, null if it is not present in the profile
      */
     Attribute getAttribute(String name);
 
     /**
-     * Return the value of the first attribute with a name starting with <code>name</code>
+     * Return a list of all the {@link Attribute} with a name starting with <code>name</code>
```
echarrod: Does it work if we add an `s` here so it's `a list of all the {@link Attribute}s`?

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/SimpleAgeVerification.java
```diff
@@ -0,0 +1,47 @@
...
+public class SimpleAgeVerification implements AgeVerification {
+
+    private final Attribute<String> derivedAttribute;
+    private final int ageVerified;
+    private final String checkPerformed;
+    private final boolean result;
+
+    public SimpleAgeVerification(Attribute<String> derivedAttribute) {
+        Validation.notNull(derivedAttribute, "derivedAttribute");
+        Validation.matchesPattern(derivedAttribute.getName(), "[^:]+:(?!.*:)[0-9]+", "attribute.name");
```
sidibos: I would have moved this process into a Processor and make this class as stupid as possible.
It should just get the data from the constructor, set them and return the AgeVerification object.

yoti-sdk-api/src/main/java/com/yoti/api/client/Profile.java
bucky-boy: Seems to display ok

yoti-sdk-api/src/main/java/com/yoti/api/client/Profile.java
```diff
@@ -1,44 +1,55 @@
...
+     * @param clazz attribute type
+     * @return typed attribute value, <code>null</code> if there was no match
      */
     <T> Attribute<T> findAttributeStartingWith(String name, Class<T> clazz);
 
     /**
      * Return all attributes for the profile.
      *
-     * @return an unsorted collection of attributes
+     * @return an unsorted collection of {@link Attribute}
```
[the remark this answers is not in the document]
bucky-boy: ok

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/HumanProfileAdapter.java
bucky-boy: As far as we know, we'll only ever deal with _age_over_ + _age_under_.

The plan is to wrap the _Attribute_ object in an _AgeVerification_ object.  The _AgeVerification_ object provides additional, relevant functionality - so it is the helper.  We shouldn't need to expose another helper beyond that one.  So the processing to be done is 
1. deciding what to wrap and what to wrap it in, and 
2. producing the wrapper

Pulling logic out into dedicated components is normally nice.  It's normally more testable, you can keep things focussed.  But in these SDKs we have no DI framework, so if we split things up we have to wire all the components together ourselves.  If we do it at every opportunity we'll likely end up in a mess, and in this case for _findAllAgeVerifications()_ it seems quite reasonable to keep it inside _HumanProfileAdapter_.

The thing I don't like with the _AgeProcessor_ and _Age Under|Over Processor_ is that each Processor is responsible for identifying what it matches.  So its possible to have more than one _Processor_ claiming to be the wrapper for a given attribute, and I don't think we want to go down that road.

We could maintain a mapping of attribute derivations to the classes that should wrap them, then iterate all the attributes and check the map to see if it matches.  The map keys have to be unique, so we can be sure there will only ever be one match.  That's a perfectly valid alternative to this approach.  If we do ever get more than two _age_derivations_ we could switch to that if we think its cleaner.

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/SimpleAgeVerification.java
bucky-boy: Here, the _SimpleAgeVerification_ is just asserting that its being created with the correct information.  It's validating it's own creation / being responsible for (what will be) it's own content, which is reasonable.

It could go into a _SimpleAgeVerification_ factory, but I don't think there's enough going on here to justify creating one.  We're only creating one type of object, and we're not going to reuse a factory anywhere else.

And it's happening inside the constructor, so there's not really a sensible way to inject some other component.  You might use static code, but I'm trying to avoid that in most cases because it often gets tricky to unit test cleanly.

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/HumanProfileAdapter.java
sidibos: Ok, either way is fine by me as long as we can easily expand this functionality.
Also I do like having ageVerifications array as a map, where the key is the attribute name and the value is the ageVerification object.

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/SimpleAgeVerification.java
sidibos: Ok

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/HumanProfileAdapter.java
bucky-boy: Yep, I'm just changing that

yoti-sdk-api/src/main/java/com/yoti/api/client/AgeVerification.java
```diff
@@ -0,0 +1,36 @@
...
+ * Wraps an 'Age Verify/Condition' attribute to provide behaviour specific to verifying someone's age
+ */
+public interface AgeVerification {
+
+    /**
+     * The type of age check performed, as specified on dashboard. Currently this might be 'age_over' or 'age_under'
+     *
+     * @return The type of age check performed
+     */
+    String getCheckPerformed();
```
qzagarese: I don't really like the naming.
I'm more for something like `getCheckType()`.
I know you don't want to use enums, but I would still change the name and make it return a string that is a constant.

yoti-sdk-api/src/main/java/com/yoti/api/client/AgeVerification.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * The age that was that checked, as specified on dashboard.
+     *
+     * @return The age that was that checked
+     */
+    int getAgeVerified();
```
qzagarese: Here I would either put simply `getAge()`, because you call it in the context of an age verification,
or `getVerificationAge()`.

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/SimpleAgeVerification.java
```diff
@@ -0,0 +1,47 @@
...
+        this.derivedAttribute = derivedAttribute;
+
+        String[] split = derivedAttribute.getName().split(":");
+        this.checkPerformed = split[0];
+        this.ageVerified = Integer.parseInt(split[1]);
+        this.result = parseBoolean(derivedAttribute.getValue());
+    }
+
+    @Override
+    public int getAgeVerified() {
```
qzagarese: this will have to change accordingly to the interface definition.

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/SimpleAgeVerification.java
```diff
@@ -0,0 +1,47 @@
...
+        return ageVerified;
+    }
+
+    @Override
+    public String getCheckPerformed() {
```
qzagarese: same here

PR #91, opened 2018-11-08
 SDK-672: Start sending SDK version in the headers

bucky-boy:
We will start sending an additional header, that allows us to see which _version_ of a given SDK people are using.

I did try to find a way of pulling the .jar version in at runtime, but it doesn't work reliably when the code is not in a jar.  I figured it's not really worth it.

### DIFF
--- a/yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/YotiConstants.java
+++ b/yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/YotiConstants.java
@@ -13,11 +13,13 @@ public final class YotiConstants {
     public static final String AUTH_KEY_HEADER = "X-Yoti-Auth-Key";
     public static final String DIGEST_HEADER = "X-Yoti-Auth-Digest";
     public static final String YOTI_SDK_HEADER = "X-Yoti-SDK";
+    public static final String YOTI_SDK_VERSION_HEADER = YOTI_SDK_HEADER + "-Version";
 
     public static final String CONTENT_TYPE = "Content-Type";
     public static final String CONTENT_TYPE_JSON = "application/json";
 
     public static final String JAVA = "Java";
+    public static final String SDK_VERSION = JAVA + "-2.2.0";
     public static final String BOUNCY_CASTLE_PROVIDER = "BC";
     public static final String SIGNATURE_ALGORITHM = "SHA256withRSA";
     public static final String ASYMMETRIC_CIPHER = "RSA/NONE/PKCS1Padding";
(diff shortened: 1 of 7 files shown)


yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/YotiConstants.java
```diff
@@ -13,11 +13,13 @@ private YotiConstants() {}
...
     public static final String DIGEST_HEADER = "X-Yoti-Auth-Digest";
     public static final String YOTI_SDK_HEADER = "X-Yoti-SDK";
+    public static final String YOTI_SDK_VERSION_HEADER = YOTI_SDK_HEADER + "-Version";
 
     public static final String CONTENT_TYPE = "Content-Type";
     public static final String CONTENT_TYPE_JSON = "application/json";
 
     public static final String JAVA = "Java";
+    public static final String SDK_VERSION = JAVA + "-2.2.0";
     public static final String BOUNCY_CASTLE_PROVIDER = "BC";
```
irotech: u'd use this one instead BouncyCastleProvider.PROVIDER_NAME

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/YotiConstants.java
```diff
@@ -13,11 +13,13 @@ private YotiConstants() {}
     public static final String AUTH_KEY_HEADER = "X-Yoti-Auth-Key";
     public static final String DIGEST_HEADER = "X-Yoti-Auth-Digest";
     public static final String YOTI_SDK_HEADER = "X-Yoti-SDK";
+    public static final String YOTI_SDK_VERSION_HEADER = YOTI_SDK_HEADER + "-Version";
```
irotech: has this '-' as prefix a meaning?

echarrod: > has this '-' as prefix a meaning?

It looks like it's just a separator from `SDK` in `X-Yoti-SDK`

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/YotiConstants.java
bucky-boy: Good spot, I'll include that with the Dynamic QR Codes changes

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/YotiConstants.java
bucky-boy: Yes, it's just a separator for readability

PR #93, opened 2018-11-09
SDK-496: Apply New Demo Design

bucky-boy:
Finally, a real change!

This PR applies the new demo design from the static project.  Details are in the Jira ticket.  Note that there's an open question about _application_logo_.

The static files are all lifted straight from the static project.  No changes.

The new design works with a slightly different model for displaying attributes.  I've done that mapping in the Java spring controller rather than in the page (much easier).  The other difference is that the approach here will display _any_ attribute returned in the profile, whereas the static design _suggests_ we'd only use the named ones we know about now.  We don't want our demos to do that.


yoti-sdk-spring-boot-example/src/main/resources/templates/index.html
```diff
@@ -1,22 +1,62 @@
...
+
+    <body class="yoti-body">
+
+        <main>
+            <section class="yoti-top-section">
+                <div class="yoti-logo-section">
+                    <img class="yoti-logo-image" th:src="@{/static/assets/logo.png}" th:srcset="@{/static/assets/logo@2x.png 2x}" alt="Yoti"/>
+                </div>
+
+                <h1 class="yoti-top-header">We now accept Yoti</h1>
```
sidibos: This wording is not the same as the original template which says 'Business Name now uses Yoti'.
Are we changing this?

yoti-sdk-spring-boot-example/src/main/resources/templates/profile.html
```diff
@@ -1,88 +1,99 @@
...
+                </div>
+
+                <div class="yoti-attribute-list">
+
+                    <div th:each="attribute : ${displayAttributes}" class="yoti-attribute-list-item">
+
+                        <div class="yoti-attribute-name">
+                            <div class="yoti-attribute-name-cell">
+                                <i th:class="${attribute.icon}" />
+                                <span class="yoti-attribute-name-cell-text" th:text="${attribute.displayName}" />
```
sidibos: Will this work if the attribute is not ticked in the Dashboard' scenario tab?

yoti-sdk-spring-boot-example/src/main/resources/templates/index.html
bucky-boy: Well spotted.
Yes, we are changing it.  Agree with Kiran it's not worth making people configure a 'Business Name' in the app for the sake of displaying it here.  
Noted in SDK-493.

yoti-sdk-spring-boot-example/src/main/resources/templates/profile.html
bucky-boy: This will only try to display attributes that were returned in the profile.

yoti-sdk-spring-boot-example/src/main/java/com/yoti/api/examples/springboot/YotiLoginController.java
```diff
@@ -57,44 +65,82 @@ public String home(final Model model) {
...
     }
 
-    private void addAttributeToModel(Model model, Attribute attribute) {
-        if (attribute != null) {
-            String attributeName = CaseFormat.LOWER_UNDERSCORE.to(CaseFormat.LOWER_CAMEL, attribute.getName());
-            model.addAttribute(attributeName, attribute.getValue());
-            model.addAttribute(attributeName + "Sources", attribute.getSources());
-            model.addAttribute(attributeName + "Verifiers", attribute.getVerifiers());
+    private DisplayAttribute mapToDisplayAttribute(Attribute attribute) {
+        switch (attribute.getName()) {
```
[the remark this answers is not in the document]
bucky-boy: We also need _icon_.  So we'd need to keep 2x values in the Map, which Java doesn't do OOTB

yoti-sdk-spring-boot-example/src/main/resources/templates/profile.html
```diff
@@ -1,88 +1,99 @@
...
+                                <div class="yoti-attribute-anchors -subtype" th:text="${anchor.subType}" />
+                            </div>
+
+                            <div th:each="anchor : ${attribute.attribute.verifiers}">
+                                <div class="yoti-attribute-anchors -s-v">Verifier</div>
+                                <div class="yoti-attribute-anchors -value" th:text="${anchor.value}" />
+                                <div class="yoti-attribute-anchors -subtype" th:text="${anchor.subType}" />
+                            </div>
+                        </div>
+
```
echarrod: Did you add these blank lines manually or is it your linter?

bucky-boy: I would have done that

echarrod: I know you're matching how the tags were opened, but I think it's tidier without line breaks, especially when you don't have line breaks after some of the div tags. But no problem if you want to keep it like this!

bucky-boy: I prefer the line breaks, keeping divs in clear blocks.  I'm going to leave it...

bucky-boy:
Agreed with @sidibos I am good to merge this

PR #96, opened 2018-12-10
SDK-706: Add support for parentRememberMeId

bucky-boy:
This adds support for the _parentRememberMeId_ property, handling it in the same way as _rememberMeId_.
1. We expect to receive a _byte[]_ on the _receipt_, named _parent_remember_me_id_.
2. We convert that to a Base64 encoded string, and return it on the _ActivityDetails_ object.

The only difference is that _parentRememberMeId_ may be null, whereas _rememberMeId_ may not.


yoti-sdk-api/src/main/java/com/yoti/api/client/ActivityDetails.java
```diff
@@ -39,6 +39,14 @@
      */
     String getRememberMeId();
 
+    /**
+     * Return the parent-remember-me ID, which is a unique, stable identifier for a user in the context of an organisation/user
```
echarrod: Remember Me ID is `rememberMeId` in the developer docs, so do you think we should format it as either `ParentRememberMeID` or `Parent-Remember-Me-ID` here? Looks a bit disjointed currently

yoti-sdk-api/src/main/java/com/yoti/api/client/ActivityDetails.java
```diff
@@ -39,6 +39,14 @@
...
+     * interaction. You can be use it to identify returning users.  This value is consistent for a given user across different apps
+     * belonging to a single organisation.
+     *
+     * @return user ID.
```
echarrod:
```suggestion
     * @return Parent Remember Me ID.
```

Looks like the other `@return`s in this file need updating too

yoti-sdk-api/src/main/java/com/yoti/api/client/ActivityDetails.java
bucky-boy: Ok, renamed to rememberMeId & parentRememberMeId

yoti-sdk-api/src/main/java/com/yoti/api/client/ActivityDetails.java
bucky-boy: Well spotted, fixed them

yoti-sdk-api/src/main/java/com/yoti/api/client/ActivityDetails.java
```diff
@@ -25,32 +25,40 @@
...
      * interaction. You can be use it to identify returning users. Note that it is different for the same user in
      * different applications.
      *
-     * @return user ID.
+     * @return rememberMeId
      */
     String getRememberMeId();
 
+    /**
+     * Return the parentRememberMeId, which is a unique, stable identifier for a user in the context of an organisation/user
```
qzagarese: just `organisation` instead of `organisation/user`

bucky-boy: > in the context of an organisation/user interaction.

if I take out _user_, that becomes

> in the context of an organisation interaction.

I don't think that makes much sense, an interaction would normally involve two parties.

echarrod: We could probably drop `interaction` too couldn't we? So it's `a unique, stable identifier for a user in the context of an organisation.`

bucky-boy: Ok, changed.  So for _getRememberMeId_ we now have:
> Return the rememberMeId, which is a unique, stable identifier for a user in the context of an application.  You can use it to identify returning users. This value will be different for the same user in different applications.

And for _getParentRememberMeId_:
> Return the parentRememberMeId, which is a unique, stable identifier for a user in the context of an organisation.  You can use it to identify returning users.  This value is consistent for a given user across different apps belonging to a single organisation.

PR #99, opened 2018-12-18
Release 2.2

bucky-boy:
This PR is for merging the entire _DEVELOPMENT_ branch into master.  So among stuff, this release will include:
- Sandbox module
- New AgeVerification object
- Client for Dynamic QR Codes
- New protobufs
- New demo app
- parentRememberMeId

All of these changes have been reviewed + approved previously, except for
- the last one - bumping the version numbers 2.2.0 (because we don't check that in to the _DEVELOPMENT_ branch).
- I've fixed the new _AgeVerification_ example in the README


yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/YotiConstants.java
```diff
@@ -6,19 +6,20 @@ private YotiConstants() {}
...
     public static final String DIGEST_HEADER = "X-Yoti-Auth-Digest";
     public static final String YOTI_SDK_HEADER = "X-Yoti-SDK";
+    public static final String YOTI_SDK_VERSION_HEADER = YOTI_SDK_HEADER + "-Version";
 
     public static final String CONTENT_TYPE = "Content-Type";
     public static final String CONTENT_TYPE_JSON = "application/json";
 
     public static final String JAVA = "Java";
-    public static final String BOUNCY_CASTLE_PROVIDER = "BC";
+    public static final String SDK_VERSION = JAVA + "-2.2.0";
```
sidibos: Why does it say `-2.2.0` is this a Java thing?

bucky-boy: We are using _major-minor-patch_ for our versioning.  Not likely we'll make much use of the patch release, but it's there if we need it.

So here the SDK version will be _Java-2.2.0_.

echarrod: Do we need to have `Java` in front of it? There's the separate `X-Yoti-SDK` header which will identify it as Java right? in [.NET](https://github.com/getyoti/yoti-dotnet-sdk/commit/7c22325549c581f4a7f075dc00e1235330a80a40) I just have it as the version without the identifier in front of it

bucky-boy: On the one hand, no we don't need it since we have the other header.

On the other hand, if we find we're looking at reports that are only based on one header value, not both, then knowing that it's **Java**-2.2.0 will be quite helpful.

sidibos: I thought these two headers `X-Yoti-SDK` and `X-Yoti-SDK-Version` always go together.

PR #103, opened 2019-03-07
SDK-757: Add support for document_images attribute

bucky-boy:
1. I noticed a mistake with the mock-maker-inline file required by Mockito, it was in the wrong place.  I fixed it.

2. Added support for _document_images_.  I can't see a better way to support Multi-Value than this.
- We can parse any multi-value, including nested multi-values
- If the attribute _name_ is _document-images_ then we strip out any value that is not an _Image_.
-- So we do _not_ support nested multi-values for this particular attribute
-- We only support JPEG and PNG.  If another, unrecognised image type is added we fallback to handling it as a _String_, which will then be stripped out because it's not an _Image_
- Therefore the new _getDocumentImages_ can safely return _Attribute<List\<Image>>_
- A call to _getAttribute(name)_ has to return _Attribute\<List>_
-- Is very nearly the same as _Attribute<List\<Object>>_ (they can safely cast it themselves, though it doesn't give them anything)
-- We cannot give nice OOTB behaviour for any _List\<T>_ - Java generics won't stretch that far.

Note that in Java, an Object can tell you it's underlying type.  So if you do...
```
Object object = new JpegAttributeValue(new byte[5]);
System.out.println(object.getClass());
```
... it will output...
```
class com.yoti.api.client.spi.remote.JpegAttributeValue
```

In the case of our new multi-value, when people come to assign to a variable they either have to
1. Use _switch_ (or _if_) on the object.getClass()
2. Use _instanceof_
3. Know the expected generic type in advance, and cast to that.  To be fair, that has always been the expected behaviour of _getAttribute(name)_


yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/AttributeConverter.java
```diff
@@ -47,32 +50,59 @@ static final AttributeConverter newInstance() {
     }
 
     private Object convertValueFromProto(AttrProto.Attribute attribute) throws ParseException, IOException {
-        switch (attribute.getContentType()) {
+        return convertValue(attribute.getContentType(), attribute.getValue());
```
echarrod: This method doesn't really add anything does it? Now that we're just calling the method below

davidgrayston: I think this is so that the switch logic in `convertValue()` can be reused by `convertMultiValue()` without breaking existing calls to `convertValueFromProto()`, which expects `AttrProto.Attribute`.

`convertMultiValue()` needs to pass `ContentTypeProto.ContentType` to `convertValue()` (not `AttrProto.Attribute`)

bucky-boy: Not much, no.

I just thought maybe leaving it might make it clearer that there are two different ways in to _convertValue_.  Maybe just more obvious / more readable.  This class is starting to get larger, there's more to follow.

Can remove it if you want.

davidgrayston: Is the alternative to change the arguments of `convertValueFromProto()` and change line 44:
```java
Object value = convertValueFromProto(attribute);
```
to:
```java
Object value = convertValueFromProto(attribute.getContentType(), attribute.getValue());
```
Instead?

echarrod: > Is the alternative to change the arguments of `convertValueFromProto()` and change line 44:
> 
> ```java
> Object value = convertValueFromProto(attribute);
> ```
> 
> to:
> 
> ```java
> Object value = convertValueFromProto(attribute.getContentType(), attribute.getValue());
> ```
> 
> Instead?

That's what I was thinking. We don't need the whole attribute object since we're only using `contentType` and `value`. I think it would be cleaner if we removed this method, but it's up to you @bucky-boy!

PR #108, opened 2019-03-18
SDK-757: Log a warning whenever we fail to return a MultiValue entry

bucky-boy:


yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/AttributeConverter.java
```diff
@@ -87,17 +87,19 @@ private Object convertSpecialType(AttrProto.Attribute attribute, Object value) t
...
-                return filterForImages((List<Object>) value);
+                return filterForImages((List<Object>) value, Image.class, AttributeConstants.HumanProfileAttributes.DOCUMENT_IMAGES);
             default:
                 return value;
         }
     }
 
-    private List<Image> filterForImages(List<Object> values) {
-        List<Image> list = new ArrayList<>();
+    private <T> List<T> filterForImages(List<Object> values, Class<T> clazz, String attributeName) {
```
echarrod: Since this is now generic, should we call it something like `filterForType`?

bucky-boy: Well spotted, thank you.  Yes we should.

PR #109, opened 2019-03-19
SDK-864: Add support for parsing INT attribute values

bucky-boy:
We just parse the value from _byte[]_ to _String_ to _Integer_.

Note that if we get an _empty_ value (protobuf equivalent to null), this will fail.  That's what I'd expect.  If they mean to give us 0 (zero), they should give us 0.  Giving us _empty_ is no good.


bucky-boy:
> Presume we're using `Attribute<Integer>` rather than `Attribute<int>` so clients can have methods like `.toString()` on the resulting value?

We have to use a proper _Class_ in a generic type.  _Integer_ is a class, but _int_ is a primitive - so it's not allowed.

PR #113, opened 2019-07-29
SDK-1010: Update references of Yoti Dashboard -> Yoti Hub

MrBurtyyy:
- All references in the codebase have been updated to use Yoti Hub/Hub instead of Yoti Dashboard/Dashboard
- Example spring boot project configuration file was updated to reference hub rather than dashboard


README.md
```diff
@@ -265,11 +265,11 @@ Yoti will provide a boolean result on the following checks:
 * Fraud list - Verify against  US Social Security Administration Fraud (SSN Fraud) list
 * Watch list - Verify against watch lists from the Office of Foreign Assets Control
 
-To use this functionality you must ensure your application is assigned to your Organisation in the Yoti Dashboard - please see [here](https://www.yoti.com/developers/documentation/#1-creating-an-organisation) for further information.
+To use this functionality you must ensure your application is assigned to your Organisation in the Yoti Hub - please see [here](https://www.yoti.com/developers/documentation/#1-creating-an-organisation) for further information.
```
emmas-yoti: I was told the URI for the documentation should be changed to developers.yoti.com. So this should be:
https://developers.yoti.com/yoti-app-integration/yoti-app-integration#step-1-creating-an-organisation

echarrod: We have SDK-1081 to do that now. Since there's a small chance the links could change, I think it's best to leave this as is for now, and then we'll update the remaining ones in one go, once they've been confirmed 👍

yoti-sdk-spring-boot-example/README.md
```diff
@@ -21,10 +21,10 @@ Before you start, you'll need to create an Application in [Dashboard](https://ww
...
 
 ## Building your example server-app
-1. In the [Yoti Dashboard](https://www.yoti.com/dashboard/applications) set the application domain of your app to `https://localhost:8443/`. Note that your endpoint must be accessible from the machine that is displaying the QR code.
-1. Still in the dashboard, set the scenario callback URL to `/login`. 
-1. Edit the [resources/application.yml](src/main/resources/application.yml) and replace the `yoti-client-sdk-id-from-dashboard` value with the `Yoti client SDK ID` you can find in Dashboard.
-1. Download your Application's key pair from Yoti-Dashboard and copy it to `resources/app-keypair.pem`.
+1. In the [Yoti Hub](https://hub.yoti.com) set the application domain of your app to `https://localhost:8443/`. Note that your endpoint must be accessible from the machine that is displaying the QR code.
+1. Still in the Hub, set the scenario callback URL to `/login`. 
+1. Edit the [resources/application.yml](src/main/resources/application.yml) and replace the `yoti-client-sdk-id-from-hub` value with the `Yoti client SDK ID` you can find in Yoti Hub.
+1. Download your Application's key pair from Yoti-Hub and copy it to `resources/app-keypair.pem`.
```
echarrod: I think we can remove the `-` here:
```suggestion
1. Download your Application's key pair from Yoti Hub and copy it to `resources/app-keypair.pem`.
```

yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/DynamicScenario.java
```diff
@@ -11,7 +11,7 @@
 public interface DynamicScenario {
 
     /**
-     * The device's callback endpoint. Must be a URL relative to the Application Domain specified in your Dashboard
+     * The device's callback endpoint. Must be a URL relative to the Application Domain specified in your Hub
```
davidgrayston: Does this need to be "Yoti Hub"?

echarrod: Yes, I think we might as well put Yoti in to make sure we're clear. Also I'm not sure if `in Yoti Hub` is better than `in your Yoti Hub`?

```suggestion
     * The device's callback endpoint. Must be a URL relative to the Application Domain specified in Yoti Hub
```

PR #114, opened 2019-07-30
SDK-1095: Fix Auth Types

MrBurtyyy:
- Update the logic in withSelfieAuthorisation and withPinAuthorisation to allow the user to set to false
- Write unit tests to support chaining these methods gives the desired outcome (as well as checking no exceptions are raised if set to false without first turning it on)

### DIFF
--- a/yoti-sdk-impl/src/main/java/com/yoti/api/client/shareurl/policy/SimpleDynamicPolicyBuilder.java
+++ b/yoti-sdk-impl/src/main/java/com/yoti/api/client/shareurl/policy/SimpleDynamicPolicyBuilder.java
@@ -186,6 +186,8 @@ public class SimpleDynamicPolicyBuilder extends DynamicPolicyBuilder {
     public DynamicPolicyBuilder withSelfieAuthorisation(boolean enabled) {
         if (enabled) {
             return withWantedAuthType(SELFIE_AUTH_TYPE);
+        } else {
+            this.wantedAuthTypes.remove(SELFIE_AUTH_TYPE);
         }
         return this;
     }
@@ -200,6 +202,8 @@ public class SimpleDynamicPolicyBuilder extends DynamicPolicyBuilder {
     public DynamicPolicyBuilder withPinAuthorisation(boolean enabled) {
         if (enabled) {
             return withWantedAuthType(PIN_AUTH_TYPE);
+        } else {
+            this.wantedAuthTypes.remove(PIN_AUTH_TYPE);
         }
         return this;
     }
--- a/yoti-sdk-impl/src/main/java/com/yoti/api/client/shareurl/policy/SimpleDynamicPolicyBuilder.java
+++ b/yoti-sdk-impl/src/main/java/com/yoti/api/client/shareurl/policy/SimpleDynamicPolicyBuilder.java
@@ -184,12 +184,7 @@ public class SimpleDynamicPolicyBuilder extends DynamicPolicyBuilder {
 
     @Override
     public DynamicPolicyBuilder withSelfieAuthorisation(boolean enabled) {
-        if (enabled) {
-            return withWantedAuthType(SELFIE_AUTH_TYPE);
-        } else {
-            this.wantedAuthTypes.remove(SELFIE_AUTH_TYPE);
-        }
-        return this;
+        return this.withWantedAuthType(SELFIE_AUTH_TYPE, enabled);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


yoti-sdk-impl/src/test/java/com/yoti/api/client/shareurl/policy/SimpleDynamicPolicyBuilderTest.java
```diff
@@ -192,6 +193,48 @@ public void buildWithRememberMeFlags() {
...
     }
 
+    @Test
+    public void buildWithSelfieAuthorisationEnabledThenDisabled() {
+        DynamicPolicy result = new SimpleDynamicPolicyBuilder()
+                .withSelfieAuthorisation(true)
+                .withSelfieAuthorisation(false)
+                .build();
+
+        assertThat(result.getWantedAuthTypes(), not(hasItem(1)));
```
echarrod: Is it possible to refer to the constants for the auth types here, instead of `1`/`2`?

MrBurtyyy: If we make the constants public rather than private, yes @echarrod . Again this is something we're seeing in the Node SDK where these weren't available to anything else originally so I'll speak to @davidgrayston tomorrow about it.

yoti-sdk-impl/src/test/java/com/yoti/api/client/shareurl/policy/SimpleDynamicPolicyBuilderTest.java
```diff
@@ -192,6 +193,48 @@ public void buildWithRememberMeFlags() {
...
+        DynamicPolicy result = new SimpleDynamicPolicyBuilder()
+                .withPinAuthorisation(false)
+                .withPinAuthorisation(true)
+                .build();
+
+        assertThat(result.getWantedAuthTypes(), hasItem(2));
+    }
+
+
+
```
echarrod: I think there might be too many line breaks here. 
Also please could you add tests to check it handles being initialised with just `.withPinAuthorisation(false)` or `.withPinAuthorisation(false)` 👍

MrBurtyyy: Fixed in [022188f]

yoti-sdk-impl/src/main/java/com/yoti/api/client/shareurl/policy/SimpleDynamicPolicyBuilder.java
```diff
@@ -200,6 +202,8 @@ public DynamicPolicyBuilder withPinAuthorisation() {
     public DynamicPolicyBuilder withPinAuthorisation(boolean enabled) {
         if (enabled) {
             return withWantedAuthType(PIN_AUTH_TYPE);
+        } else {
+            this.wantedAuthTypes.remove(PIN_AUTH_TYPE);
```
davidgrayston: Can we move this into `withWantedAuthType()` as per _Node_ and _.NET_ _(in a non-breaking way)_? This will allow any auth type to be added/removed

yoti-sdk-impl/src/test/java/com/yoti/api/client/shareurl/policy/SimpleDynamicPolicyBuilderTest.java
davidgrayston: As discussed, we could define constants for the test to make this more readable

yoti-sdk-impl/src/test/java/com/yoti/api/client/shareurl/policy/SimpleDynamicPolicyBuilderTest.java
```diff
@@ -192,6 +193,64 @@ public void buildWithRememberMeFlags() {
...
+        assertThat(result.getWantedAuthTypes(), hasItem(2));
+    }
+
+    @Test
+    public void buildsWithSelfieAuthorisationDisabled() {
+        DynamicPolicy result = new SimpleDynamicPolicyBuilder()
+                .withSelfieAuthorisation(false)
+                .build();
+
+        assertThat(result.getWantedAuthTypes(), hasItem(1));
```
davidgrayston: Should this be `not(hasItem(1))`?

echarrod: Once we add Travis tests, these can automatically be picked up 😃 👍

PR #117, opened 2019-07-31
SDK-1034: Add unknown anchor test

MrBurtyyy:
- Added a test to ensure the SDK correctly handles unknown anchors as per the example from the Ruby SDK.

### DIFF
--- a/yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/AnchorConverterTest.java
+++ b/yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/AnchorConverterTest.java
@@ -39,6 +39,7 @@ public class AnchorConverterTest {
     private static final String PASSPORT_SOURCE_TYPE = "PASSPORT";
     private static final String DRIVING_LICENCE_SOURCE_TYPE = "DRIVING_LICENCE";
     private static final String YOTI_ADMIN_VERIFIER_TYPE = "YOTI_ADMIN";
+    private static final String YOTI_UNKNOWN_ISSUER = "CN=document-registration-server";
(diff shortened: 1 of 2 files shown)


yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/AnchorConverterTest.java
```diff
@@ -39,6 +39,7 @@
     private static final String PASSPORT_SOURCE_TYPE = "PASSPORT";
     private static final String DRIVING_LICENCE_SOURCE_TYPE = "DRIVING_LICENCE";
     private static final String YOTI_ADMIN_VERIFIER_TYPE = "YOTI_ADMIN";
+    private static final String YOTI_UNKNOWN_ISSUER = "CN=document-registration-server";
```
davidgrayston: Could this be in `TestAnchors.java`?

MrBurtyyy: Yep it absolutely could do - I have put it in the test file to follow convention of the other constants.  We could rename it to just UNKNOWN_ISSUER to follow the convention used.

davidgrayston: Ah yes, I see what you mean - maybe it should be `DOCUMENT_REGISTRATION_ISSUER` and remain where it is - @echarrod thoughts?

echarrod: I think naming it `DOCUMENT_REGISTRATION_ISSUER` and leaving it where it is seems sensible.

PR #121, opened 2019-08-02
SDK-882: Add travis file for build verification

MrBurtyyy:
- Added .travis.yml file to run build verification within travis-ci.com
- Updated README to include build status badge (for master branch)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,49 @@
+language: java
+
+cache:
+  directories:
+    $HOME/.m2
+
+git:
+  depth: 1
+
+jobs:
+  include:
+    - state: Test
+      jdk: oraclejdk8
+      install: ./mvnw install -DskipTests=true -Dmaven.javadoc.skip=true -B -V -Dgpg.skip=true -Ddependency-check.skip=true -Dhttps.protocols=TLSv1.2
+      script:
+        - ./mvnw test -B
+    - &test-compatability
+      stage: Compatability
+      jdk: oraclejdk8
+      os: linux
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,48 @@
...
+
+cache:
+  directories:
+    $HOME/.m2
+
+env:
+  - JAVA_TOOL_OPTIONS=-Dhttps.protocols=TLSv1.2
+
+git:
+  depth: 1
```
echarrod: Although having a git depth of 1 saves a tiny amount of time, I think @davidgrayston and I discussed that we would rather have an update depth of 3, so that we can better track our failing metrics (i.e. we won't get as many `failed` builds because of another commit superseding it). We still have a depth of 1 in some travis files, but the plan is to change to 3 I think.
```suggestion
  depth: 3
```

echarrod:
Just the minor comment about the git depth 👌 

Since this involved a lot of fiddling around with the commits, would it be possible to do an interactive rebase and `fixup` the commits so that they only have 2/3/4 commits? Just so the commit history isn't quite as long, and if someone is wanting to look through the commit history for the changes made, they won't see all of these. 👍

MrBurtyyy:
> Just the minor comment about the git depth 👌
> 
> Since this involved a lot of fiddling around with the commits, would it be possible to do an interactive rebase and `fixup` the commits so that they only have 2/3/4 commits? Just so the commit history isn't quite as long, and if someone is wanting to look through the commit history for the changes made, they won't see all of these. 👍

Fixed the commits into larger commits where important changes were made!

PR #123, opened 2019-08-05
SDK-997: Update attribute text

MrBurtyyy:
- Updated comments to include possible values for SOURCES and VERIFIERS

### DIFF
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/Anchor.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/Anchor.java
@@ -10,8 +10,7 @@ import java.util.List;
 public interface Anchor {
 
     /**
-     * The type of Anchor - most likely either but it's possible that new Anchor types will be added in future
-     * "SOURCE" or "VERIFIER"
+     * The type of Anchor - most likely "SOURCE" or "VERIFIER" but it's possible that new Anchor types will be added in future
      * @return String, naming the type of the anchor
      */
     String getType();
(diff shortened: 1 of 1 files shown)


yoti-sdk-api/src/main/java/com/yoti/api/client/Anchor.java
```diff
@@ -10,23 +10,27 @@
 public interface Anchor {
 
     /**
-     * The type of Anchor - most likely either SOURCE or VERIFIER, but it's possible that new Anchor types will be added in future
-     *
+     * The type of Anchor - most likely either but it's possible that new Anchor types will be added in future
+     * "SOURCE" or "VERIFIER"
```
davidgrayston: Should `"SOURCE" or "VERIFIER"` be after the sentence? Not sure that `most likely either but` reads well?

yoti-sdk-parent/pom.xml
```diff
@@ -102,7 +102,7 @@
     <!-- sdk dependency versions -->
     <slf4j.version>1.7.26</slf4j.version>
     <bouncy.castle.version>1.60</bouncy.castle.version>
-    <jackson.version>2.9.8</jackson.version>
+    <jackson.version>2.9.9.2</jackson.version>
```
emmas-yoti: Is this deliberate?

MrBurtyyy: I think this came from a mistake in an interactive rebase (last commit in development before this branch was created), and this change should already be in the development branch so hopefully should be able to be merged without problem.

MrBurtyyy: Yep this is already in the development branch [see here](https://github.com/getyoti/yoti-java-sdk/commit/82f2ca7c1616291504221d8956a90801e63d44f8).  I can try to find a way to remove this from the PR, or we can just merge as is if it doesn't cause a problem

PR #128, opened 2019-08-06
Release 2.4.0

MrBurtyyy:
- [x] Waiting on #130 

### Added
- `DynamicPolicyBuilder.withWantedAuthType(int wantedAuthType, boolean enabled)`
- `DynamicPolicyBuilder.withWantedAttribute(boolean optional, String name)`

### Changed
- Updated dependency versions for:
  - `jackson-databind`
  - `mockito`
  - `maven-javadoc-plugin`

### Notes
- `Yoti Dashboard` has been renamed to `Yoti Hub` with documentation reflecting this
- Example Spring Boot application has been updated to include dynamic share


yoti-sdk-api/src/main/java/com/yoti/api/client/Anchor.java
```diff
@@ -10,23 +10,25 @@
...
+     * The type of Anchor - most likely "SOURCE" or "VERIFIER" but it's possible that new Anchor types will be added in future
      * @return String, naming the type of the anchor
      */
     String getType();
 
     /**
-     * For SOURCE Anchors, describes how the attribute value was obtained, for example this might be 'OCR' or 'NFC'
-     * For VERIFIER Anchors, describes the verification method used for the attribute
+     * <pre>
+     * For SOURCE Anchors, describes how the attribute value was obtained, will be one of: ['USER_PROVIDED', 'PASSPORT', 'DRIVING_LICENSE', 'PASSCARD', 'NATIONAL_ID']
```
echarrod: Sorry Alex, I missed this the first time. Could we change the phrasing to say that the options listed are _among_ the possible options (rather than `will be one of`), so that we're not saying that the ones listed are the only options. We also can include `YOTI_AI` as an option now. (see .[NET](https://github.com/getyoti/yoti-dotnet-sdk/pull/105/files))

yoti-sdk-api/src/main/java/com/yoti/api/client/Anchor.java
```diff
@@ -10,23 +10,25 @@
...
      *
+     * For VERIFIER Anchors, describes the verification method used for the attribute, will be one of: ['YOTI_ADMIN', 'YOTI_IDENTITY', 'YOTI_OTP', 'PASSPORT_NFC_SIGNATURE', 'ISSUING_AUTHORITY', 'ISSUING_AUTHORITY_PKI']
```
echarrod: Same here

yoti-sdk-api/src/main/java/com/yoti/api/client/Anchor.java
```diff
@@ -10,23 +10,25 @@
...
+     * </pre>
      * @return String, further detailing how the attribute value was obtained
      */
     String getSubType();
 
     /**
      * Identifies the provider that either sourced or verified the attribute value.
-     * The range of possible values is not limited.  For a SOURCE anchor, expect values like PASSPORT, DRIVING_LICENSE...
+     * The range of possible values is not limited.
+     * For example, for a passport, this would either be "NFC" or "OCR"
```
echarrod: I also think these are the wrong way round - `For example, for a passport, this would either be "NFC" or "OCR"` is describing getSubType, and the range of possible Source/Verifier options listed above should be part of this `getValue` method 👍

yoti-sdk-api/src/main/java/com/yoti/api/client/Anchor.java
```diff
@@ -16,20 +16,20 @@
...
      * @return String, further detailing how the attribute value was obtained
      */
     String getSubType();
 
     /**
-     * Identifies the provider that either sourced or verified the attribute value.
-     * The range of possible values is not limited.
-     * For example, for a passport, this would either be "NFC" or "OCR"
+     * <pre>
+     * For SOURCE Anchors, describes how the attribute value was obtained, options listed are among possible values: ['USER_PROVIDED', 'PASSPORT', 'DRIVING_LICENSE', 'PASSCARD', 'NATIONAL_ID']
```
echarrod: This phrasing here might be a little ambiguous, this might denote that "the possible values are in the following list". I was thinking `among the possible options are:`. Maybe I've been looking at it for too long though...@davidgrayston what do you think?

PR #130, opened 2019-08-07
SDK-1109: Add ability to add WantedAttribute by name

MrBurtyyy:
- To bring the Java SDK inline with others, expose the functionality to add a WantedAttribute by name through the DynamicPolicyBuilder


yoti-sdk-impl/src/main/java/com/yoti/api/client/shareurl/policy/SimpleDynamicPolicyBuilder.java
```diff
@@ -29,7 +29,8 @@ public DynamicPolicyBuilder withWantedAttribute(WantedAttribute wantedAttribute)
         return this;
     }
 
-    private DynamicPolicyBuilder withWantedAttribute(boolean optional, String name) {
+    @Override
+    public DynamicPolicyBuilder withWantedAttribute(boolean optional, String name) {
```
davidgrayston: We should add test coverage for this public method

PR #132, opened 2019-10-03
SDK-1162: Signed Request Builder

MrBurtyyy:
## Added

* `SignedRequestBuilder` and `SignedRequest` to assist users with creating and executing a signed request
* `UnsignedPathFactory` for returning internal SDK paths without signature query parameters

## Changed

* Internal SDK now uses SignedRequestBuilder for making calls to Yoti API's

### DIFF
--- a/yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/aml/RemoteAmlService.java
+++ b/yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/aml/RemoteAmlService.java
@@ -1,60 +1,44 @@
 package com.yoti.api.client.spi.remote.call.aml;
 
-import static java.net.HttpURLConnection.HTTP_BAD_REQUEST;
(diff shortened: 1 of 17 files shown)


yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/aml/RemoteAmlService.java
```diff
@@ -11,50 +11,43 @@
...
+        );
     }
 
-    public RemoteAmlService(ResourceFetcher resourceFetcher,
-            PathFactory pathFactory,
-            HeadersFactory headersFactory,
+    public RemoteAmlService(
+            UnsignedPathFactory unsignedPathFactory,
             ObjectMapper objectMapper,
-            SignedMessageFactory signedMessageFactory) {
```
echarrod: Is this not a breaking change?

PR #137, opened 2019-10-15
SDK-1233: Attribute Issuance Details

MrBurtyyy:
## Added

* `ActivityDetails`
  * `getExtraData()` - Returns the extra data associated with the receipt
* `ExtraData`
  * `getAttributeIssuanceDetails()` - Returns the issuance details from extra data
* `AttributeIssuanceDetails`
  * `getExpiryDate()` - The expiry date of the token
  * `getToken()` - The token to be used by the organisation
  * `getIssuingAttributes()` - The list of attributes to be issued

## Deprecated

* `SimpleActivityDetails` 
  * Deprecated old constructor in favour of new one that accepts `ExtraData`


yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/ActivityDetailsFactory.java
```diff
@@ -40,11 +46,25 @@ ActivityDetails create(Receipt receipt, PrivateKey privateKey) throws ProfileExc
...
+
         String rememberMeId = parseRememberMeId(receipt.getRememberMeId());
         String parentRememberMeId = parseRememberMeId(receipt.getParentRememberMeId());
         Date timestamp = parseTimestamp(receipt.getTimestamp());
 
-        return new SimpleActivityDetails(rememberMeId, parentRememberMeId, userProfile, applicationProfile, timestamp, receipt.getReceiptId());
+        return new SimpleActivityDetails(rememberMeId, parentRememberMeId, userProfile, applicationProfile, extraData, timestamp, receipt.getReceiptId());
+    }
+
+    private ExtraData parseExtraData(byte[] extraDataBytes) throws ProfileException {
```
echarrod: I know we have ProfileExceptions being thrown elsewhere in ActivityDetails, but I don't think it makes sense to throw a ProfileException here, since ExtraData is not part of the profile, is there a suitable alternative?

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/DataEntryConverterTest.java
```diff
@@ -0,0 +1,82 @@
...
+    @Test(expected = ExtraDataException.class)
+    public void shouldWrapInvalidProtocolBufferException() throws Exception {
+        InvalidProtocolBufferException invalidProtocolBufferException = new InvalidProtocolBufferException("Some Exception");
+
+        when(dataEntryMock.getType()).thenReturn(DataEntryProto.DataEntry.Type.THIRD_PARTY_ATTRIBUTE);
+        when(dataEntryMock.getValue()).thenReturn(SOME_DATA_ENTRY_BYTE_STRING);
+        when(thirdPartyAttributeConverterMock.parseThirdPartyAttribute(SOME_DATA_ENTRY_BYTE_STRING.toByteArray())).thenThrow(invalidProtocolBufferException);
+
+        dataEntryConverter.convertDataEntry(dataEntryMock);
+    }
```
echarrod: Should there be any assertions here? Should we be asserting that the exception is of the expected type?

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/DataEntryConverterTest.java
```diff
@@ -0,0 +1,82 @@
...
+    @Test(expected = ExtraDataException.class)
+    public void shouldWrapParseException() throws Exception {
+        ParseException parseException = new ParseException("Some exception", 0);
+
+        when(dataEntryMock.getType()).thenReturn(DataEntryProto.DataEntry.Type.THIRD_PARTY_ATTRIBUTE);
+        when(dataEntryMock.getValue()).thenReturn(SOME_DATA_ENTRY_BYTE_STRING);
+        when(thirdPartyAttributeConverterMock.parseThirdPartyAttribute(SOME_DATA_ENTRY_BYTE_STRING.toByteArray())).thenThrow(parseException);
+
+        dataEntryConverter.convertDataEntry(dataEntryMock);
+    }
```
echarrod: Same here - checking that this _isn't_ a ParseException?

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/credentials/ThirdPartyAttributeConverterTest.java
```diff
@@ -0,0 +1,89 @@
...
+        assertThat(issuingAttributes, hasSize(1));
+        assertThat(issuanceToken, is("someIssuanceToken"));
+
+        assertEquals(2019, dateTimeValue.getDate().getYear());
+        assertEquals(10, dateTimeValue.getDate().getMonth());
+        assertEquals(15, dateTimeValue.getDate().getDay());
+
+        assertEquals(22, dateTimeValue.getTime().getHour());
+        assertEquals(04, dateTimeValue.getTime().getMinute());
+        assertEquals(05, dateTimeValue.getTime().getSecond());
```
echarrod: If there isn't already a method to check that the datetime matches, do you think it would be worth adding one? So we don't have to use 6 lines to check

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/ActivityDetailsFactory.java
MrBurtyyy: As we talked about, in the next major we are going to review the use of checked exceptions so we decided to wrap the ExtraDataException into a profile exception to not make it a breaking change for users.

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/DataEntryConverterTest.java
MrBurtyyy: No - we're just checking that it managed to create the data entry successfully and didn't throw any exceptions.  The method `throws Exception` so that the test will fail if any exceptions are thrown.

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/DataEntryConverterTest.java
MrBurtyyy: Same again, in the `@Test` annotation we're expecting an `ExtraDataException` to be thrown

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/credentials/ThirdPartyAttributeConverterTest.java
MrBurtyyy: When I looked through the code, I couldn't find a function that did this.  We could probably create one though that has a method signature something around the lines of:

`assertDateTime(DateValue check, int year, int month, int day, int hour, int minute, int second)`.

Thoughts?

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/credentials/ThirdPartyAttributeConverter.java
```diff
@@ -0,0 +1,80 @@
...
+import com.yoti.api.client.spi.remote.proto.IssuingAttributesProto;
+import com.yoti.api.client.spi.remote.proto.ThirdPartyAttributeProto;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.*;
+
+public class ThirdPartyAttributeConverter {
+
+    private static final String RFC3339_PATTERN = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'";
```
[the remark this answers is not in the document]
MrBurtyyy: Yep - might move this into the static constants class we have

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/credentials/ThirdPartyAttributeConverterTest.java
MrBurtyyy: Fixed in 5a78f1f

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/credentials/ThirdPartyAttributeConverter.java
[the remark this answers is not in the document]
MrBurtyyy: Fixed in https://github.com/getyoti/yoti-java-sdk/pull/137/commits/082b4fbb46e6fd283c6d2fc4c5da6dae39325e2a

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/credentials/ThirdPartyAttributeConverterTest.java
echarrod: 👌

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/DataEntryConverterTest.java
MrBurtyyy: This test is just to make sure that the exception is wrapped into an `ExtraDataException`.  It's more generic than the one you have described above to make sure that any exceptions of a certain type are wrapped, not in a specific place which is what you have described does.  We can add in some more specific tests though 👍

echarrod: Ah ok, that's fine, I was just worried that it was just checking for a general Exception, but didn't see the annotation above where we check for the specific type. This is fine as is then.

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/DataEntryConverterTest.java
echarrod: Same for this one, all good then

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/ActivityDetailsFactory.java
```diff
@@ -17,14 +17,11 @@
 import java.text.SimpleDateFormat;
 import java.util.Date;
 
-import static com.yoti.api.client.spi.remote.call.YotiConstants.DEFAULT_CHARSET;
-import static com.yoti.api.client.spi.remote.call.YotiConstants.SYMMETRIC_CIPHER;
+import static com.yoti.api.client.spi.remote.call.YotiConstants.*;
```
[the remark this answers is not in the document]
MrBurtyyy: This was done by the autoformatter in Intellij - I can put it back but if someone uses autoformat it will just go back to this.  @bucky-boy what's the general rule of thumb around using * for static imports?

MrBurtyyy:
I've updated this so that we have `AttributeIssuanceDetails` now, and instead of storing a list of Strings for the issuing attributes, we have moved to using a list of `Definition` as these might get added to in the future

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/ActivityDetailsFactoryTest.java
```diff
@@ -126,6 +127,26 @@ public void shouldFailWithInvalidTimestamp() {
...
+        when(extraDataConverterMock.read(eq(EXTRA_DATA_CONTENT))).thenThrow(extraDataException);
+
+        try {
+            testObj.create(receipt, keyPair.getPrivate());
+        } catch (ProfileException e) {
+            assertThat(e.getMessage(), containsString("Failed to load extra data from receipt"));
+            assertTrue(e.getCause() instanceof ExtraDataException);
+            return;
+        }
+        fail("Expected an Exception");
```
echarrod: Is this exception being caught above this? We want to make sure that an exception with ExtraData isn't affecting the rest of the receipt parsing, so ideally we'd check that this error is thrown, logged, and the parsing continues.

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/ThirdPartyAttributeConverterTest.java
```diff
@@ -0,0 +1,125 @@
...
+        assertThat(attributeIssuanceDetails.getIssuingAttributes().get(0).getName(), is("SOME_DEFINITION"));
+        assertThat(attributeIssuanceDetails.getIssuingAttributes().get(1).getName(), is("SOME_OTHER_DEFINITION"));
+    }
+
+    @Test
+    public void shouldParseThirdPartyAttributeWithoutToken() throws Exception {
+        byte[] thirdPartyAttribute = buildThirdPartyAttribute("", null).toByteArray();
+        AttributeIssuanceDetails attributeIssuanceDetails = thirdPartyAttributeConverter.parseThirdPartyAttribute(thirdPartyAttribute);
+
+        assertThat(attributeIssuanceDetails.getToken(), is(""));
```
echarrod: I think we should be returning null for `attributeIssuanceDetails` here, so clients know for certain that something is amiss. Can we also check the log/error message too in this scenario?

MrBurtyyy: I spoke to @bucky-boy about checking for logs, and he said we should only do it if there was a business critical need for the log (i.e. a service that was looking for the log and triggered something else).

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/ThirdPartyAttributeConverter.java
```diff
@@ -0,0 +1,107 @@
...
+
+        String token = thirdPartyAttribute.getIssuanceToken().toStringUtf8();
+        DateTime expiryDate = issuingAttributes.getExpiryDate();
+        List<AttributeDefinition> attributeDefinitions = issuingAttributes.getAttributeDefinitions();
+
+        if (token == null || token.isEmpty()) {
+            LOG.error("Failed to retrieve token from ThirdPartyAttribute");
+        }
+        if (expiryDate == null) {
+            LOG.error("Failed to retrieve expiryDate from ThirdPartyAttribute");
```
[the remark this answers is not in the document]
MrBurtyyy: Yep sounds good to me - I'll log in the parsing method rather than here (which I think is actually already done)

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/ActivityDetailsFactoryTest.java
MrBurtyyy: It is currently caught above, but we could change this to just catch it and log it (not affecting the rest of the parsing)

MrBurtyyy: Fixed in https://github.com/getyoti/yoti-java-sdk/pull/137/commits/aaa16277bd2bf529518c5465a2f8ade07a78c641

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/ThirdPartyAttributeConverter.java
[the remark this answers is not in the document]
MrBurtyyy: Fixed in https://github.com/getyoti/yoti-java-sdk/pull/137/commits/39e63eafb4369a2a2b9ba7aeaca89fd7db039194

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/ThirdPartyAttributeConverterTest.java
MrBurtyyy: Fixed in https://github.com/getyoti/yoti-java-sdk/pull/137/commits/4a8dcceb86480d32d8a4c5b5161c6f9d6caee28a

PR #138, opened 2019-10-24
SDK-1127: Source constraints

MrBurtyyy:
## Added

* Support for source constraints for `WantedAttribute` when using Dynamic Scenarios

### DIFF
--- a/yoti-sdk-impl/src/main/java/com/yoti/api/client/shareurl/constraint/SimpleSourceConstraintBuilder.java
+++ b/yoti-sdk-impl/src/main/java/com/yoti/api/client/shareurl/constraint/SimpleSourceConstraintBuilder.java
@@ -1,0 +1,39 @@
+package com.yoti.api.client.shareurl.constraint;
+
+import com.yoti.api.client.shareurl.policy.WantedAnchor;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class SimpleSourceConstraintBuilder implements SourceConstraintBuilder {
+
+    private List<WantedAnchor> wantedAnchors;
+
+    private boolean softPreference;
+
+    SimpleSourceConstraintBuilder() {
+        this.wantedAnchors = new ArrayList<>();
+        this.softPreference = false;
+    }
+
+    @Override
+    public SourceConstraintBuilder withWantedAnchor(WantedAnchor wantedAnchor) {
+        this.wantedAnchors.add(wantedAnchor);
+        return this;
+    }
+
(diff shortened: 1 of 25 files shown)


yoti-sdk-impl/src/main/java/com/yoti/api/client/shareurl/constraint/SimpleSourceConstraintBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public SourceConstraintBuilder withSoftPreference(boolean softPreference) {
+        this.softPreference = softPreference;
+        return this;
+    }
+
+    @Override
+    public SourceConstraint build() {
+        return new SimpleSourceConstraint(
+                Collections.unmodifiableList(wantedAnchors),
```
echarrod: Not sure if we need to enforce this being unmodifiable here - this is just for the client building the request to be sent to us, I don't think we mind if it's altered before it's sent to us (unlike the profile objects that we return), would you agree?

yoti-sdk-impl/src/main/java/com/yoti/api/client/shareurl/policy/SimpleWantedAttribute.java
```diff
@@ -17,15 +20,24 @@
...
+    @JsonProperty("constraints")
+    private final List<Constraint> constraints;
+
+    SimpleWantedAttribute(String name, String derivation, boolean optional, boolean acceptSelfAsserted, List<Constraint> constraints) {
         this.name = name;
         this.derivation = derivation;
         this.optional = optional;
+        this.acceptSelfAsserted = acceptSelfAsserted;
+
+        // TODO: Check with Mike if this is ok?
```
echarrod: TODO

MrBurtyyy: Fixed in https://github.com/getyoti/yoti-java-sdk/pull/138/commits/8bde085bd4cc694e43381e6a29a1ccbf62662768

yoti-sdk-impl/src/main/java/com/yoti/api/client/shareurl/constraint/SimpleSourceConstraintBuilder.java
MrBurtyyy: As Lists are passed by reference, would prefer to keep it unmodifiable once the WantedAttribute has been built 👍

PR #141, opened 2019-10-25
SDK-1148: Multiple of the same attribute with different constraints

MrBurtyyy:
- [x] Merge #138 into `DEVELOPMENT`

## Added

* `Profile`
  * Added methods for getting an attribute by exact name, as the old `getAttribute` method used the same filtering functionality as `findAttributeByName` which checked if the attribute name started with the name supplied.
  * Changed the way Attributes are stored within Profile to use a list rather than a map.  This is to support receiving multiple attributes with the same name (i.e. document_images with different constraints)
* `DynamicPolicyBuilder`
  * The key of each wanted attribute is now modified if the attribute has a list of constraints attached to it.  This is to support the requesting of the same attribute multiple times with different constraints in Dynamic Scenarios

## Deprecated

* `Profile`
  * The old `getAttribute` methods have been deprecated in favour of using `getAttributeByName` using an equality match.

### DIFF
--- a/yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/SimpleProfileTest.java
+++ b/yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/SimpleProfileTest.java
@@ -242,7 +242,7 @@ public class SimpleProfileTest {
     }
 
     @Test
-    public void getAttributeByName_shouldReturnAllMatchingAttributesWhenMultipleWithSameName() {
+    public void getAttributesByName_shouldReturnAllMatchingAttributesWhenMultipleWithSameName() {
         List<Attribute<?>> attributeList = new ArrayList<>();
         attributeList.add(createAttribute("some_attribute", "firstValue"));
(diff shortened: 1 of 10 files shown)


yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/SimpleProfileTest.java
```diff
@@ -165,9 +228,41 @@ public void findAttributeStartingWith_shouldReturnTypedValueForMatchingKey() {
...
+
+        SimpleProfile profile = new SimpleProfile(attributeList);
+
+        Attribute<String> result = profile.getAttributeByName("some_attribute", String.class);
+
+        assertEquals(result.getValue(), "firstValue");
+    }
+
+    @Test
+    public void getAttributeByName_shouldReturnAllMatchingAttributesWhenMultipleWithSameName() {
```
davidgrayston:
```suggestion
    public void getAttributesByName_shouldReturnAllMatchingAttributesWhenMultipleWithSameName() {
```

PR #142, opened 2019-10-28
SDK-1264: Third Party Attribute Extension

MrBurtyyy:
- [x] Merge #137 into `DEVELOPMENT`

### DIFF
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/extension/ThirdPartyAttributeContent.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/extension/ThirdPartyAttributeContent.java
@@ -1,0 +1,32 @@
+package com.yoti.api.client.shareurl.extension;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.yoti.api.client.AttributeDefinition;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.List;
+import java.util.TimeZone;
+
+public class ThirdPartyAttributeContent {
+
+    private final Date expiryDate;
+
+    @JsonProperty("definitions")
+    private final List<AttributeDefinition> definitions;
+
+    public ThirdPartyAttributeContent(Date expiryDate, List<AttributeDefinition> definitions) {
+        this.expiryDate = expiryDate;
+        this.definitions = definitions;
+    }
+
+    @JsonProperty("expiry_date")
+    public String getExpiryDate() {
+        SimpleDateFormat rfcDateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+        return rfcDateFormat.format(expiryDate.getTime());
+    }
+
+    public List<AttributeDefinition> getDefinitions() {
+        return definitions;
+    }
+}
--- a/yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/share/ThirdPartyAttributeIssuingRequestBuilder.java
+++ b/yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/share/ThirdPartyAttributeIssuingRequestBuilder.java
@@ -1,0 +1,73 @@
+package com.yoti.api.client.spi.remote.call.share;
+
+import com.yoti.api.client.AttributeDefinition;
+import com.yoti.api.client.spi.remote.IssuingAttribute;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ThirdPartyAttributeIssuingRequestBuilder {
+
+    private String issuanceToken;
+
+    private List<IssuingAttribute> issuingAttributes;
+
+    public ThirdPartyAttributeIssuingRequestBuilder() {
+        this.issuingAttributes = new ArrayList<>();
+    }
+
+    /**
+     * Sets the issuance token to be used with the {@link ThirdPartyAttributeIssuingRequest}
+     *
+     * @param issuanceToken the issuance token
+     * @return the builder
+     */
+    public ThirdPartyAttributeIssuingRequestBuilder withIssuanceToken(String issuanceToken) {
+        this.issuanceToken = issuanceToken;
+        return this;
+    }
+
+    /**
+     * Adds an {@link IssuingAttribute} to be used in the {@link ThirdPartyAttributeIssuingRequest}
+     *
+     * @param definition the attribute definition
+     * @param value      the value of the {@link IssuingAttribute}
+     * @return the builder
+     */
+    public ThirdPartyAttributeIssuingRequestBuilder withIssuingAttribute(AttributeDefinition definition, String value) {
+        IssuingAttribute issuingAttribute = new IssuingAttribute(definition, value);
+        return withIssuingAttribute(issuingAttribute);
+    }
+
+    /**
+     * Adds an {@link IssuingAttribute} to be used in the {@link ThirdPartyAttributeIssuingRequest}
+     *
+     * @param issuingAttribute the issuing attribute
+     * @return the builder
+     */
+    public ThirdPartyAttributeIssuingRequestBuilder withIssuingAttribute(IssuingAttribute issuingAttribute) {
+        this.issuingAttributes.add(issuingAttribute);
+        return this;
+    }
+
+    /**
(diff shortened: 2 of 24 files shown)


MrBurtyyy:
This has become quite a complex pull request, with this branch and #137 becoming very intertwined.  It might be worth just reviewing this PR @echarrod @davidgrayston as it's encapsulates everything required for third party attribute flow.

yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/extension/ThirdPartyAttributeContent.java
```diff
@@ -0,0 +1,32 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.List;
+import java.util.TimeZone;
+
+public class ThirdPartyAttributeContent {
+
+    private final Date expiryDate;
+
+    @JsonProperty("definitions")
```
echarrod: Should we put this annotation on the public property, not the private one?

MrBurtyyy: Don't think so - generally you put the annotation on the field declaration, unless you need it to be different (i.e. in the case of the expiry date I want it to be a string of a certain format)

echarrod: I think the JSON serializer in .NET (newtonsoft) takes all of the public properties/fields of a class and serializes them. So this would get added to the payload twice. But as long as it doesn't do that in Java, then this is fine!

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/share/ThirdPartyAttributeIssuingRequestBuilder.java
```diff
@@ -0,0 +1,43 @@
...
+        IssuingAttribute issuingAttribute = new IssuingAttribute(definition, value);
+        return withIssuingAttribute(issuingAttribute);
+    }
+
+    public ThirdPartyAttributeIssuingRequestBuilder withIssuingAttribute(IssuingAttribute issuingAttribute) {
+        this.issuingAttributes.add(issuingAttribute);
+        return this;
+    }
+
+    public ThirdPartyAttributeIssuingRequestBuilder withIssuingAttributes(List<IssuingAttribute> issuingAttributes) {
```
echarrod: Would be good if we annotated this to make sure users know that any previously added attributes will be overridden. Then if we do that, I think we should add annotations to the other methods too.

yoti-sdk-impl/src/test/java/com/yoti/api/client/shareurl/extension/SimpleThirdPartyAttributeExtensionBuilderTest.java
```diff
@@ -0,0 +1,138 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+public class SimpleThirdPartyAttributeExtensionBuilderTest {
+
+    public static final Date SOME_DATE = new Date();
+    public static final String SOME_DEFINITION = "com.thirdparty.id";
```
echarrod: These don't need to be public do they?

yoti-sdk-impl/src/test/java/com/yoti/api/client/shareurl/extension/SimpleThirdPartyAttributeExtensionBuilderTest.java
```diff
@@ -0,0 +1,138 @@
...
+        assertEquals(ExtensionConstants.THIRD_PARTY_ATTRIBUTE, extension.getType());
+
+        List<AttributeDefinition> definitions = extension.getContent().getDefinitions();
+        assertThat(definitions.size(), is(2));
+        assertThat(definitions.get(0).getName(), is("firstDefinition"));
+        assertThat(definitions.get(1).getName(), is("secondDefinition"));
+    }
+
+    @Test
+    public void doTest() throws Exception {
```
echarrod: I think this test maybe isn't finished, or can be removed 👍

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/share/ThirdPartyAttributeIssuingRequestBuilder.java
MrBurtyyy: Fixed in https://github.com/getyoti/yoti-java-sdk/pull/142/commits/9f85e1478a0c1ea43cbf1b24bd1bf5d3040a8691

yoti-sdk-impl/src/test/java/com/yoti/api/client/shareurl/extension/SimpleThirdPartyAttributeExtensionBuilderTest.java
MrBurtyyy: Nope, good spot.  Will change these 👍

yoti-sdk-impl/src/test/java/com/yoti/api/client/shareurl/extension/SimpleThirdPartyAttributeExtensionBuilderTest.java
MrBurtyyy: Fixed in https://github.com/getyoti/yoti-java-sdk/pull/142/commits/04fe75becb7960203ed6e759fb81251c010bf64d

yoti-sdk-impl/src/test/java/com/yoti/api/client/shareurl/extension/SimpleThirdPartyAttributeExtensionBuilderTest.java
MrBurtyyy: Fixed in https://github.com/getyoti/yoti-java-sdk/pull/142/commits/04fe75becb7960203ed6e759fb81251c010bf64d

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/ExtraDataConverterTest.java
```diff
@@ -0,0 +1,142 @@
...
+                buildDataEntry(THIRD_PARTY_ATTRIBUTE, ByteString.copyFromUtf8("someValue")),
+                buildDataEntry(THIRD_PARTY_ATTRIBUTE, ByteString.copyFromUtf8("someOtherValue"))
+        );
+
+        ExtraData extraData = extraDataConverter.read(extraDataBytes);
+        AttributeIssuanceDetails attributeIssuanceDetails = extraData.getAttributeIssuanceDetails();
+
+        assertNotNull(attributeIssuanceDetails);
+        assertEquals(attributeIssuanceDetails.getToken(), "someValue");
+        assertNotEquals(attributeIssuanceDetails.getToken(), "someOtherToken");
```
davidgrayston: Could we remove this line?

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/DataEntryConverter.java
```diff
@@ -0,0 +1,54 @@
...
+        return entry;
+    }
+
+    private Object convertValue(DataEntryProto.DataEntry.Type dataEntryType, ByteString value) throws
+            InvalidProtocolBufferException, ExtraDataException {
+        switch (dataEntryType) {
+            case THIRD_PARTY_ATTRIBUTE:
+                return thirdPartyAttributeConverter.parseThirdPartyAttribute(value.toByteArray());
+            default:
+                LOG.warn("Unsupported data entry, skipping...");
```
echarrod: Looks like we say "skipping" here, but then when we return null to the caller, we're throwing an exception here. Do we want to return this null value in the caller, or change the error message here so it doesn't say "skipping"?

echarrod: In fact we should definitely do the former, so that if there are multiple data entries, an unsupported first one won't prevent the others from being parsed.

PR #145, opened 2019-11-14
[Revision] SDK-1148: Multiple attributes same name

MrBurtyyy:
## Changed

* `SimpleProfile` now stores the Attributes in a Map of `List<Attribute>`.

### DIFF
--- a/yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/SimpleProfile.java
+++ b/yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/SimpleProfile.java
@@ -1,17 +1,16 @@
 package com.yoti.api.client.spi.remote;
 
 import static java.util.Collections.unmodifiableList;
+import static java.util.Collections.unmodifiableMap;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
+import java.util.*;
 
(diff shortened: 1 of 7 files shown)


yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/SimpleProfile.java
```diff
@@ -89,10 +60,12 @@ public Attribute getAttributeByName(String name) {
     public <T> List<Attribute<T>> getAttributesByName(String name, Class<T> clazz) {
         ensureName(name);
         List<Attribute<T>> matches = new ArrayList<>();
-        for (Attribute<?> entry : protectedAttributes) {
-            if (entry.getName().equals(name)) {
-                Attribute<T> value = castSafely(clazz, entry);
-                matches.add(value);
+        for (Map.Entry<String, List<Attribute<?>>> entry : protectedAttributes.entrySet()) {
```
davidgrayston: Do we need to iterate here or can be access by key?

MrBurtyyy: Fixed in https://github.com/getyoti/yoti-java-sdk/pull/145/commits/535954712a20d28fcfcc17ce17d87cd8468d7302

PR #147, opened 2019-11-22
SDK-1274: Update Signed Request & Builder

MrBurtyyy:
## Added

* Support for different Content-Types that are used in the Yoti ecosystem (mainly images and binary data used in Doc Scan)

### DIFF
--- a/yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/ImageResourceFetcher.java
+++ b/yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/ImageResourceFetcher.java
@@ -1,0 +1,55 @@
+package com.yoti.api.client.spi.remote.call;
+
+import static com.yoti.api.client.spi.remote.call.YotiConstants.CONTENT_TYPE;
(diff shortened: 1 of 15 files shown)


yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/ImageResourceFetcher.java
```diff
@@ -0,0 +1,55 @@
...
+        SignedRequestResponse signedRequestResponse = rawResourceFetcher.doRequest(signedRequest);
+        String contentType = getContentType(signedRequestResponse.getResponseHeaders());
+        byte[] responseBody = signedRequestResponse.getResponseBody();
+        switch (contentType) {
+            case CONTENT_TYPE_PNG:
+                return new PngAttributeValue(responseBody);
+            case CONTENT_TYPE_JPEG:
+                return new JpegAttributeValue(responseBody);
+            default:
+                LOG.error("Failed to convert to PNG or JPEG Attribute Value");
```
echarrod: Usually I'm not a fan of error messages like this, which could forget to be added to if we add further types (e.g. "unsupported image type" might be better). But having said that, I can't foresee us adding support for any images any time soon, so I think this is fine.

PR #149, opened 2019-11-27
SDK-1281: Check for and except if wanted attribute name is empty

emmas-yoti:


yoti-sdk-impl/src/main/java/com/yoti/api/client/shareurl/policy/SimpleWantedAttribute.java
```diff
@@ -29,6 +29,10 @@
     private final List<Constraint> constraints;
 
     SimpleWantedAttribute(String name, String derivation, boolean optional, Boolean acceptSelfAsserted, List<Constraint> constraints) {
+        if (name == "") {
+            throw new IllegalArgumentException("Attribute name missing");
+        }
```
MrBurtyyy: I suggest we use the Validation module we have that gets used in the rest of the system.  Suggested change:

```java
Validation.notNullOrEmpty(name, "name");
```

Can we also add in a test to ensure the exception is thrown if a name isn't supplied?

yoti-sdk-impl/src/test/java/com/yoti/api/client/shareurl/policy/SimpleWantedAttributeBuilderTest.java
```diff
@@ -55,4 +55,9 @@ public void buildsAnAttributeWithSourceConstraint() {
         assertEquals(result.getConstraints(), constraintListMock);
     }
 
+    @Test(expected = IllegalArgumentException.class)
+    public void buildingAnAttributeWithoutNameFails() {
+        WantedAttributeBuilder.newInstance().build();
+    }
```
MrBurtyyy: Can we change this test to check that the `IllegalArgumentException` message contains `name`?  I think `builder_shouldApplyMaxLimitToDay()` in `DateValueTest.java` is a good example of how to structure it

PR #155, opened 2019-12-12
SDK-1322: Add dependabot configuration file

MrBurtyyy:
## Added

* Dependabot configuration file to change default behaviour

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,12 @@
+version: 1
+update_configs:
+  - package_manager: "java:maven"
+    directory: "/"
+    update_schedule: "live"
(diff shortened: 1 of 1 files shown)


.dependabot/config.yml
```diff
@@ -0,0 +1,12 @@
+version: 1
+update_configs:
+  - package_manager: "java:maven"
+    directory: "/"
+    update_schedule: "live"
```
echarrod:
```suggestion
    update_schedule: "daily"
```
Same here:
> The property '#/update_configs/0/update_schedule' value "live" did not match one of the following values: daily, weekly, monthly

PR #160, opened 2020-01-21
SDK-982: Doc Scan Client

MrBurtyyy:
## Added

* API support for Yoti Doc Scan services that require signed requests, including:
  * Session Creation
  * Session Retrieval
  * Session Deletion
  * Media Retrieval
  * Media Deletion
* Builders and models to assist creating payloads required for communication with Doc Scan services


yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/docs/DocScanService.java
```diff
@@ -0,0 +1,277 @@
...
+    private final UnsignedPathFactory unsignedPathFactory;
+    private final ObjectMapper objectMapper;
+
+    private final String apiUrl;
+
+    private DocScanService(UnsignedPathFactory pathFactory, ObjectMapper objectMapper) {
+        this.unsignedPathFactory = pathFactory;
+        this.objectMapper = objectMapper;
+
+        apiUrl = System.getProperty(PROPERTY_YOTI_DOCS_URL, DEFAULT_YOTI_DOCS_URL);
```
davidgrayston: Perhaps we should consider passing this into the service as configuration/parameter instead of setting globally, as it's something integrators will need to change when using the sandbox.

It also looks like this uses the same `PROPERTY_YOTI_DOCS_URL` key as the sandbox (same applies for `PROPERTY_YOTI_DOCS_URL`) - could this cause conflicts or are they encapsulated in some way?

Also related to this, perhaps we should consider a separate client for doc scan as it has a different base URL and could potentially cause complications in other SDKs that provide the base URL to the clients using constructor or setter (we'd need to pass in a base URL for each service). As an example the [AWS SDK](https://github.com/aws/aws-sdk-java) takes this approach.

Thoughts on this @echarrod ?

echarrod: I'm a fan of having separate clients to avoid any conflicts, and also make it cleaner. This will be a problem in some SDKs more than others, but it would be ideal to adopt a common approach across them all.

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/SdkConfig.java
```diff
@@ -0,0 +1,71 @@
...
+     * @return the allowed capture methods
+     */
+    String getAllowedCaptureMethods();
+
+    /**
+     * Retrieves the primary colour from the SDK configuration
+     *
+     * @return the primary colour
+     */
+    String getPrimaryColour();
```
echarrod: Was going to say we should probably be consistent with the spelling of "colour" , as we have `ApplicationProfile.application_receipt_bgcolor`, but since this is reflecting the JSON (`primary_colour`), I'm thinking it would probably be better to leave it as it is

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/session/create/check/SimpleRequestedLivenessCheckBuilder.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class SimpleRequestedLivenessCheckBuilder extends RequestedLivenessCheckBuilder {
+
+    private String livenessType;
+    private int maxRetries = 1;
+
+    @Override
+    public RequestedLivenessCheckBuilder forZoomLiveness() {
+        this.livenessType = DocScanConstants.ZOOM;
+        return this;
```
echarrod: Don't think it makes too much difference, but should we change this to:
```suggestion
        return forLivenessType(DocScanConstants.ZOOM);
```

_(says suggestions can't be used on multi-line commits, not sure if this will appear weirdly)_

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/MediaValue.java
```diff
@@ -0,0 +1,32 @@
...
+    public byte[] getContent() {
+        return content.clone();
+    }
+
+    @Override
+    public String getBase64Content() {
+        return String.format("data:%s;base64,%s", getMimeType(), toBase64String(getContent()));
+    }
+
+}
```
echarrod: Looks like this file is almost identical to ImageAttributeValue. Should we have `JpegAttributeValue` and `PngAttributeValue` extend from this, and deprecate ImageAttributeValue?

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/docs/DocScanService.java
```diff
@@ -0,0 +1,277 @@
...
+            return signedRequest.execute(SimpleCreateSessionResult.class);
+        } catch (GeneralSecurityException ex) {
+            throw new DocScanException("Error signing the request: " + ex.getMessage(), ex);
+        } catch (ResourceException ex) {
+            throw new DocScanException("Error posting the request: " + ex.getMessage(), ex);
+        } catch (IOException | URISyntaxException ex) {
+            throw new DocScanException("Error building the request: " + ex.getMessage(), ex);
+        } catch (Exception ex) {
+            throw new DocScanException("Error creating the session: " + ex.getMessage(), ex);
+        }
```
echarrod: I'm assuming we know conclusively that these exceptions can only be thrown in the respective sign/post/build/create parts of this? Will just make debugging a fair bit harder if for any reason these exceptions can be thrown from other parts of this process if we're appending `Error <x>ing the request/session`

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/DocScanClient.java
```diff
@@ -0,0 +1,130 @@
...
+    public void deleteDocScanMedia(String sessionId, String mediaId) throws DocScanException {
+        LOG.debug("Deleting media content '{}' in session '{}'", mediaId, sessionId);
+        docScanService.deleteMediaContent(sdkId, keyPair, sessionId, mediaId);
+    }
+
+    private KeyPair loadKeyPair(KeyPairSource kpSource) throws InitialisationException {
+        try {
+            LOG.debug("Loading key pair from '{}'", kpSource);
+            return kpSource.getFromStream(new KeyStreamVisitor());
+        } catch (IOException e) {
```
echarrod: Could we add a test for this exception 👍

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/util/Validation.java
```diff
@@ -23,6 +24,12 @@ public static void notNullOrEmpty(String value, String name) {
         }
     }
 
+    public static <T> void notNullOrEmpty(Collection<T> value, String name) {
```
echarrod: Can we add tests for this 👍 Doesn't look like there's any coverage at the moment

yoti-sdk-impl/src/test/java/com/yoti/api/client/spi/remote/call/docs/DocScanServiceTest.java
```diff
@@ -0,0 +1,867 @@
...
+        doReturn(signedRequestBuilderMock).when(docScanService).getSignedRequestBuilder();
+        when(objectMapperMock.writeValueAsBytes(sessionSpecMock)).thenReturn(SOME_SESSION_SPEC_BYTES);
+        when(signedRequestBuilderMock.build()).thenReturn(signedRequestMock);
+        when(signedRequestMock.execute(SimpleCreateSessionResult.class)).thenReturn(simpleCreateSessionResultMock);
+        when(unsignedPathFactoryMock.createNewYotiDocsSessionPath(SOME_APP_ID)).thenReturn(SOME_PATH);
+
+        CreateSessionResult result = docScanService.createSession(SOME_APP_ID, KEY_PAIR, sessionSpecMock);
+
+        assertThat(result, Matchers.<CreateSessionResult>is(simpleCreateSessionResultMock));
+
```
echarrod: Not sure if they should be here or somewhere else, but can we add tests for the CreateSession getters: `getClientSessionTokenTtl`, `getClientSessionToken` and `getSessionId`?

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/factory/PathFactory.java
```diff
@@ -29,6 +29,18 @@ public String createDynamicSharingPath(String appId) {
         return unsignedPathFactory.createDynamicSharingPath(appId) + createSignatureParams();
     }
 
+    public String createNewYotiDocsSessionPath(String appId) {
```
echarrod: Can we add tests for these 3 in PathFactorytest 👍

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/session/create/SimpleSdkConfigBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+        return withAllowedCaptureMethod(DocScanConstants.CAMERA);
+    }
+
+    @Override
+    public SdkConfigBuilder withAllowsCameraAndUpload() {
+        return withAllowedCaptureMethod(DocScanConstants.CAMERA_AND_UPLOAD);
+    }
+
+    @Override
+    public SdkConfigBuilder withAllowedCaptureMethod(String allowedCaptureMethod) {
```
davidgrayston: Should this be `withAllowedCaptureMethods`?

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/DocScanClient.java
```diff
@@ -0,0 +1,128 @@
...
+
+    private final String sdkId;
+    private final KeyPair keyPair;
+
+    private final DocScanService docScanService;
+
+    DocScanClient(final String sdkId,
+                  final KeyPairSource keyPairSource,
+                  DocScanService docScanService) {
+        this.sdkId = sdkId;
```
echarrod: Assuming we can't enforce the DocScanClient being initialised through the builder, do you think we should add validation for the parameters here?

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/SdkConfigBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Sets the allowed capture method to the specified value on the builder
+     *
+     * @param allowedCaptureMethod the allowed capture method
+     * @return the builder
+     */
+    public abstract SdkConfigBuilder withAllowedCaptureMethod(String allowedCaptureMethod);
+
+    /**
+     * Sets the primary colour on the builder
```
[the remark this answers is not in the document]
echarrod: Sure, that's fine, I just took that text from our Dev Docs.

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/DocScanClient.java
MrBurtyyy: If we want to make it testable, not really.  We have default access on it which means someone would have to access the class from within the same package, whereas the builder has public access so it's kind of implied...

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/task/RequestedTextExtractionTaskBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+     */
+    public abstract RequestedTextExtractionTaskBuilder withManualFallbackAlways();
+
+    /**
+     * Sets the text extraction task to fallback to a manual check if something has gone
+     * wrong with the automated task
+     *
+     * @return the builder
+     */
+    public abstract RequestedTextExtractionTaskBuilder withManualFallbackFallback();
```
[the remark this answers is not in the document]
echarrod: Do you think we could change the wording then? e.g.:
```suggestion
    public abstract RequestedTextExtractionTaskBuilder withFallbackAsManualFallback();
```

If so, don't mind if we reorder the other ones or not

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/DocScanClient.java
echarrod: Sorry, didn't realise it was internal/package-private, this is fine then

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/MediaValue.java
echarrod: Oh ok, how come it broke the unit and integration tests? I would have thought if the exposed methods are exactly the same, it would be ok

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/docs/DocScanService.java
echarrod: Ok, if we're confident that these exceptions won't be thrown at other stages then this is fine

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/task/RequestedTextExtractionTaskBuilder.java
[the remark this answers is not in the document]
davidgrayston: Does `withManualFallbackAlways()` mean it will always trigger a manual check? Could this be `withManualCheck()` or `withManualCheckAlways()`

So we'd have something like this:
- `withManualFallback()`
- `withManualCheck()` or `withManualCheckAlways()`
- `withManualFallbackNever()`

If this is directly reflecting the API, then it's fine as is.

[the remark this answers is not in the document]
davidgrayston: Or perhaps this?
- `withManualCheck()` / `withManualCheckAlways()`
- `withManualCheckFallback()`
- `withManualCheckNever()`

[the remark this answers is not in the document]
echarrod: - withManualCheckAlways()
- withManualCheckFallback()
- withManualCheckNever()

Makes sense to me, and it reflects the JSON. Unless I'm understanding it incorrectly!

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/retrieve/BreakdownResponse.java
```diff
@@ -0,0 +1,54 @@
...
+        this.subCheck = subCheck;
+        this.result = result;
+        this.details = details;
+    }
+
+    public String getSubCheck() {
+        return subCheck;
+    }
+
+    public void setSubCheck(String subCheck) {
```
davidgrayston: Are public setters needed on `com.yoti.api.client.docs.session.retrieve.*` classes?

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/retrieve/ResourceContainer.java
```diff
@@ -0,0 +1,44 @@
...
+public class ResourceContainer {
+
+    @JsonProperty("id_documents")
+    private List<DocumentResourceResponse> idDocuments;
+
+    @JsonProperty("liveness_capture")
+    private List<LivenessResourceResponse> livenessCapture;
+
+    @JsonProperty("checks")
+    private List<CheckResponse> checks;
```
davidgrayston: Based on the documentation `checks` should only be on `DocScanSession` (not in `ResourceContainer`)

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/retrieve/GeneratedCheckResponse.java
```diff
@@ -0,0 +1,23 @@
...
+    public String getId() {
+        return id;
+    }
+
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    public String getType() {
+        return DocScanConstants.ID_DOCUMENT_TEXT_DATA_CHECK;
```
davidgrayston: Should `GeneratedCheckResponse` have type `ID_DOCUMENT_TEXT_DATA_CHECK`?

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/docs/DocScanService.java
```diff
@@ -0,0 +1,277 @@
...
+     * a session, returning information about the newly created session if successful
+     *
+     * @param appId the app ID
+     * @param keyPair the {@code KeyPair}
+     * @param sessionSpec the {@code SessionSpec}
+     * @return the session creation result
+     * @throws DocScanException if there was an error whilst performing
+     *                          the request to Doc Scan systems
+     */
+    public CreateSessionResult createSession(String appId, KeyPair keyPair, SessionSpec sessionSpec) throws DocScanException {
```
echarrod: We call this method with a parameter named "sdkId", and assuming that this is correct, can we change this to sdkId too? 
```suggestion
    public CreateSessionResult createSession(String sdkId, KeyPair keyPair, SessionSpec sessionSpec) throws DocScanException {
```

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/DocScanClient.java
```diff
@@ -0,0 +1,128 @@
...
+     * on the supplied media ID
+     *
+     * @param sessionId the session ID
+     * @param mediaId   the media ID
+     * @return {@link Media} the media
+     * @throws DocScanException if an error has occurred whilst performing
+     *                          the request to the Doc Scan systems
+     */
+    public Media getDocScanMedia(String sessionId, String mediaId) throws DocScanException {
+        LOG.debug("Deleting media content '{}' in session '{}'", mediaId, sessionId);
```
echarrod:
```suggestion
        LOG.debug("Retrieving media content '{}' in session '{}'", mediaId, sessionId);
```

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/retrieve/BreakdownResponse.java
MrBurtyyy: Fixed in the latest draft

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/retrieve/ResourceContainer.java
MrBurtyyy: Fixed in the latest draft

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/DocScanService.java
```diff
@@ -0,0 +1,260 @@
...
+    /**
+     * Retrieves the current state of a given session
+     *
+     * @param appId     the app ID
+     * @param keyPair   the {@code KeyPair}
+     * @param sessionId the session ID
+     * @return the session state
+     * @throws DocScanException if there was an error
+     */
+    public SimpleGetSessionResult retrieveSession(String appId, KeyPair keyPair, String sessionId) throws DocScanException {
```
davidgrayston: Is there a reason this is `retrieveSession()`, rather than `getSession()`?
_(do other SDKs also need to be `retrieveSession()`)_

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/retrieve/CheckResponse.java
```diff
@@ -0,0 +1,21 @@
...
+
+    String getState();
+
+    List<String> getResourcesUsed();
+
+    List<? extends GeneratedMedia> getGeneratedMedia();
+
+    ReportResponse getReport();
+
+    String getCreated();
```
davidgrayston: Do these need to be dates in other SDKs?

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/DocScanClientBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Sets the application ID to be used in the {@code DocScanClient}
+     * on the builder
+     *
+     * @param appId the application id
+     * @return the builder
+     */
+    public abstract DocScanClientBuilder withApplicationId(String appId);
```
echarrod:
```suggestion
    public abstract DocScanClientBuilder withSdkId(String sdkId);
```

echarrod: (need to update the JavaDocs above too)

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/SimpleDocScanClientBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class SimpleDocScanClientBuilder extends DocScanClientBuilder {
+
+    private static final DocScanService docScanService = DocScanService.newInstance();
+
+    private String appId;
+    private KeyPairSource keyPairSource;
+
+    @Override
+    public DocScanClientBuilder withApplicationId(String appId) {
```
echarrod:
```suggestion
    public DocScanClientBuilder withSdkId(String sdkId) {
```

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/SimpleDocScanClientBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+import static com.yoti.api.client.spi.remote.util.Validation.notNull;
+import static com.yoti.api.client.spi.remote.util.Validation.notNullOrEmpty;
+
+import com.yoti.api.client.KeyPairSource;
+
+public class SimpleDocScanClientBuilder extends DocScanClientBuilder {
+
+    private static final DocScanService docScanService = DocScanService.newInstance();
+
+    private String appId;
```
echarrod:
```suggestion
    private String sdkId;
```

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/NotificationConfig.java
```diff
@@ -0,0 +1,34 @@
+package com.yoti.api.client.docs.session.create;
+
+import java.util.List;
+
+/**
+ * Configures call-back Notifications to some backend endpoint provided by the Relying Business.
+ * <p>
+ * Notifications can be configured to notified a client backend of certain events, avoiding the neeed to poll for the state of the Session.
```
echarrod:
```suggestion
 * Notifications can be configured to notify a client backend of certain events, avoiding the need to poll for the state of the Session.
```

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/DocScanClientBuilder.java
MrBurtyyy: I think this was kept as applicationId to keep it inline with the current client which has applicationId also.  Do you think it would be better to keep them the same until the major when we can also change the old client?

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/DocScanService.java
```diff
@@ -0,0 +1,260 @@
...
+
+        try {
+            SignedRequest signedRequest = getSignedRequestBuilder()
+                    .withKeyPair(keyPair)
+                    .withBaseUrl(apiUrl)
+                    .withEndpoint(path)
+                    .withHttpMethod(HTTP_GET)
+                    .build();
+            SignedRequestResponse response = signedRequest.execute();
+            return new MediaValue(findContentType(response), response.getResponseBody());
```
davidgrayston: Could this return specific media types based on mime type in other SDKs, e.g. `Yoti\Media\Image\Jpeg` in PHP, or should we standardise and always return a generic _"Media"_ object in all SDKs?

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/call/docs/DocScanService.java
MrBurtyyy: Separate clients have been created

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/retrieve/CheckResponse.java
emmas-yoti: getCreatedTime at least for self-documentation?

echarrod: Yes, we'll set this to be the native Date/DateTime objects in other SDKs (in UTC), Java is just a special case

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/DocScanService.java
MrBurtyyy: So we could do this, but I feel like we might be backing ourselves into a corner if we do this for Doc Scan where we could have many different types of Media returned.  I think for Doc Scan, returning a generic MediaValue object is the best approach as we can let them use the mimeType and do what they want with it.  Thoughts @davidgrayston ?

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/Base64.java
```diff
@@ -75,7 +75,6 @@
  *
  *
  * @author  Xueming Shen
```
emmas-yoti: Should we get rid of the Author tag at the same time?

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/DocScanService.java
davidgrayston: Happy to always return _Media_ object, just wanted to check and follow suit in other SDKs 👍

yoti-sdk-impl/src/main/java/com/yoti/api/client/spi/remote/Base64.java
MrBurtyyy: Discussed offline, keeping tag

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/DocScanClientBuilder.java
MrBurtyyy: Fixed in d1265e4

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/SimpleDocScanClientBuilder.java
MrBurtyyy: Fixed in d1265e4

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/SimpleDocScanClientBuilder.java
MrBurtyyy: Fixed in d1265e4

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/NotificationConfig.java
MrBurtyyy: Fixed in d1265e4

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/DocScanService.java
MrBurtyyy: I don't think we have to make sure they're exactly the same, as long as the name represents the same task (i.e. retrieve and get are pretty much the same thing)

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/retrieve/CheckResponse.java
davidgrayston: I believe it's `getCreated()` to mirror the JSON property - probably neater to keep it like this and change to a date in the major release. It could have a comment explaining that it's a date string?

MrBurtyyy: Fixed in 9063944 @davidgrayston @emmas-yoti

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/DocScanClientBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Sets the SDK ID to be used in the {@code DocScanClient}
+     * on the builder
+     *
+     * @param sdkId the sdk id
+     * @return the builder
+     */
+    public abstract DocScanClientBuilder withSdkId(String sdkId);
```
echarrod: @jlmartyres do you think clients will understand this is "_Client_ SDK ID", or do you think we need to be explicit and call it `withClientSdkId`?

jlmartyres: I'd go with withClientSdkId, we refer to it similarly on Modal and it's how it's referenced in the hub too

MrBurtyyy: Fixed in 1203d1f

PR #161, opened 2020-01-21
SDK-1394: Add sonarcloud analysis

MrBurtyyy:
## Added

* Sonarcloud static code analysis

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,0 +1,9 @@
+sonar.host.url = https://sonarcloud.io
+sonar.organization = getyoti
+sonar.projectKey = getyoti:java
+sonar.projectName = Yoti Java SDK
+sonar.language = java
+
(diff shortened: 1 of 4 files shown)


sonar-project.properties
```diff
@@ -0,0 +1,9 @@
+sonar.host.url = https://sonarcloud.io
+sonar.organization = getyoti
+sonar.projectKey = getyoti:java
+sonar.projectName = Yoti Java SDK
```
echarrod: I don't mind which one we go with, but would be good to be consistent with the .NET project name. That is currently `dotnet-sdk`, I would be inclined to not have "Yoti" in the name, as `getyoti` or `Yoti Ltd` seems to mostly be present whenever we're looking at it:
image
Having said that, it also looks nicer to have ".NET SDK" and "Java SDK" rather that "dotnet-sdk" and "java-sdk", so we could go with that, and I'll update .NET too?

MrBurtyyy: Yeah makes sense - I'll update it to "Java SDK"

MrBurtyyy: Fixed in ab3013b

PR #179, opened 2020-02-20
SDK-1462: Add Doc Scan support for multiple documents and filters

MrBurtyyy:
## Added

* `DocScanClient`
  * `getSupportedDocuments()` - retrieves the list of supported documents within Doc Scan, returned as `SupportedDocumentResponse`

- `SessionSpec`
  - Can now have an optional list of `RequiredDocument` allowing relying businesses to specify certain documents required by users
- `RequiredDocument`
  - `RequiredIdentityDocument`
- `RequiredDocumentFilter`
  - `OrthogonalRestrictionsFilter`
    * `CountryRestriction`
    * `TypeRestriction`
  * `DocumentRestrictionsFilter`
    * `DocumentRestriction`

### DIFF
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/filters/DocumentFilterBuilderFactory.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/filters/DocumentFilterBuilderFactory.java
@@ -1,0 +1,25 @@
+package com.yoti.api.client.docs.session.create.filters;
+
+import java.util.ServiceLoader;
+
+public abstract class DocumentFilterBuilderFactory {
+
+    /**
+     * Returns a new instance of a {@link DocumentFilterBuilderFactory} using
+     * Java's service loader to find the first available implementation
+     *
+     * @return the factory
+     */
+    public static DocumentFilterBuilderFactory newInstance() {
+        ServiceLoader<DocumentFilterBuilderFactory> serviceLoader = ServiceLoader.load(DocumentFilterBuilderFactory.class);
+        if (!serviceLoader.iterator().hasNext()) {
+            throw new IllegalStateException("Cannot find any implementation of " + DocumentFilterBuilderFactory.class.getSimpleName());
+        }
+        return serviceLoader.iterator().next();
+    }
+
+    public abstract OrthogonalRestrictionsFilterBuilder forOrthogonalRestrictionsFilter();
+
+    public abstract DocumentRestrictionsFilterBuilder forDocumentRestrictionsFilter();
+
+}
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/filters/DocumentRestrictionsFilterBuilder.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/filters/DocumentRestrictionsFilterBuilder.java
@@ -1,0 +1,17 @@
+package com.yoti.api.client.docs.session.create.filters;
+
+import java.util.List;
+
+public interface DocumentRestrictionsFilterBuilder {
+
+    DocumentRestrictionsFilterBuilder forWhitelist();
+
+    DocumentRestrictionsFilterBuilder forBlacklist();
+
+    DocumentRestrictionsFilterBuilder withDocumentRestriction(List<String> countryCodes, List<String> documentTypes);
+
+    DocumentRestrictionsFilterBuilder withDocumentRestriction(DocumentRestriction documentRestriction);
+
+    DocumentRestrictionsFilter build();
+
+}
--- a/yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/session/create/filters/SimpleDocumentRestriction.java
+++ b/yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/session/create/filters/SimpleDocumentRestriction.java
@@ -1,0 +1,32 @@
+package com.yoti.api.client.docs.session.create.filters;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+class SimpleDocumentRestriction implements DocumentRestriction {
+
+    @JsonProperty("country_codes")
+    private List<String> countryCodes;
+
+    @JsonProperty("document_types")
+    private List<String> documentTypes;
+
+    SimpleDocumentRestriction(List<String> countryCodes, List<String> documentTypes) {
+        this.countryCodes = countryCodes;
+        this.documentTypes = documentTypes;
+    }
+
+    @Override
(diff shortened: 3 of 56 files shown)


yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/session/create/filters/document/SimpleDocumentRestriction.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.List;
+
+class SimpleDocumentRestriction implements DocumentRestriction {
+
+    @JsonProperty("country_codes")
+    private List<String> countryCodes;
+
+    @JsonProperty("document_types")
+    private List<String> documentTypes;
```
davidgrayston, line 13: I believe these should be `null` (or excluded) when the list is empty

davidgrayston: I've just referred to the Docs and it looks like empty arrays are excluded with `@JsonInclude(JsonInclude.Include.NON_EMPTY)` annotation

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/filters/RequiredDocumentBuilder.java
```diff
@@ -0,0 +1,9 @@
+package com.yoti.api.client.docs.session.create.filters;
+
+public interface RequiredDocumentBuilder<Builder extends RequiredDocumentBuilder<Builder, Filter>, Filter extends RequiredDocumentFilter> {
+
+    Builder withFilter(Filter filter);
```
davidgrayston, line 5: Should this method be removed from this interface now that it's ID document specific?

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/session/create/filters/SimpleRequiredDocumentFilterBuilderFactory.java
```diff
@@ -0,0 +1,18 @@
+package com.yoti.api.client.docs.session.create.filters;
+
+import com.yoti.api.client.docs.session.create.filters.document.DocumentRestrictionsFilterBuilder;
+import com.yoti.api.client.docs.session.create.filters.orthogonal.OrthogonalRestrictionsFilterBuilder;
+
+public class SimpleRequiredDocumentFilterBuilderFactory extends RequiredDocumentFilterBuilderFactory {
+
+    @Override
+    public OrthogonalRestrictionsFilterBuilder createOrthogonalFilter() {
```
davidgrayston, line 9:
```suggestion
    public OrthogonalRestrictionsFilterBuilder createOrthogonalRestrictionsFilter() {
```

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/filters/RequiredDocumentBuilder.java
bucky-boy: yes.  on it.

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/filters/DocumentRestrictionsFilterBuilder.java
```diff
@@ -0,0 +1,17 @@
...
+
+import java.util.List;
+
+public interface DocumentRestrictionsFilterBuilder {
+
+    DocumentRestrictionsFilterBuilder forWhitelist();
+
+    DocumentRestrictionsFilterBuilder forBlacklist();
+
+    DocumentRestrictionsFilterBuilder withDocumentRestriction(List<String> countryCodes, List<String> documentTypes);
```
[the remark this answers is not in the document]
echarrod: I can't really think of a fitting name for the `withDocumentRestriction(List<String> countryCodes, List<String> documentTypes);` method. 
So unless we can think of a good name, I'm happy to just have one of these for the languages where method overloading isn't supported.

Do we think the 1st one would be preferable? Since the object is quite simple, and building it doesn't gain much for us?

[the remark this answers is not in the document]
davidgrayston: I think `withDocumentRestriction(DocumentRestriction documentRestriction)` is preferable - this is what I've changed it to in PHP/Node - I can re-revert this?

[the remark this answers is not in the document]
davidgrayston: Discussed offline and agreed overloading in languages that can makes sense, but use `withDocumentRestriction(DocumentRestriction documentRestriction)` as the preferred method

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/session/create/filters/SimpleDocumentRestriction.java
```diff
@@ -0,0 +1,32 @@
+package com.yoti.api.client.docs.session.create.filters;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
```
davidgrayston, line 8:
```suggestion
@JsonInclude(JsonInclude.Include.NON_NULL)
```
(or remove?)

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/filters/DocumentFilterBuilderFactory.java
```diff
@@ -0,0 +1,25 @@
...
+     */
+    public static DocumentFilterBuilderFactory newInstance() {
+        ServiceLoader<DocumentFilterBuilderFactory> serviceLoader = ServiceLoader.load(DocumentFilterBuilderFactory.class);
+        if (!serviceLoader.iterator().hasNext()) {
+            throw new IllegalStateException("Cannot find any implementation of " + DocumentFilterBuilderFactory.class.getSimpleName());
+        }
+        return serviceLoader.iterator().next();
+    }
+
+    public abstract OrthogonalRestrictionsFilterBuilder forOrthogonalFilter();
```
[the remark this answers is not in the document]
bucky-boy: I'm easy about that one.  Not sure the extra word adds much, it reads ok without it.  On the hand it would be more technically correct

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/session/create/filters/SimpleDocumentRestriction.java
bucky-boy: Should be NON_NULL, I'll change it

PR #190, opened 2020-05-15
SDK-1539: Doc Scan Sandbox SDK

MrBurtyyy:
## Added

* Support for Yoti Doc Scan Sandbox
* New client `DocScanSandboxClient` which can be used to set the response configuration for a given session, or default response for an application.


yoti-sdk-sandbox/src/test/java/com/yoti/api/client/sandbox/docs/DocScanSandboxClientTest.java
```diff
@@ -0,0 +1,75 @@
...
+//        doReturn(signedRequestBuilderMock).when(docScanSandboxClient).getSignedRequestBuilder();
+//
+//        try {
+//            docScanSandboxClient.configureSessionResponse("someSessionId", responseConfigMock);
+//        } catch (SandboxException ex) {
+//            assertThat(ex.getMessage(), containsString("Some error"));
+//            return;
+//        }
+//        fail("Expected an exception");
+    }
```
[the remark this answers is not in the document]
MrBurtyyy: Fixed in 6dfcfca

yoti-sdk-sandbox/src/main/java/com/yoti/api/client/sandbox/docs/request/check/SandboxTextDataCheck.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import com.yoti.api.client.sandbox.docs.request.SandboxDocumentFilter;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public class SandboxTextDataCheck extends SandboxDocumentCheck {
```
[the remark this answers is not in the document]
MrBurtyyy: Fixed in 6dfcfca

yoti-sdk-sandbox/src/main/java/com/yoti/api/client/sandbox/docs/request/SandboxCheckReports.java
```diff
@@ -0,0 +1,118 @@
...
+
+        private List<SandboxDocumentAuthenticityCheck> documentAuthenticityCheck = new ArrayList<>();
+
+        private List<SandboxLivenessCheck> livenessCheck = new ArrayList<>();
+
+        private List<SandboxFaceMatchCheck> documentFaceMatchCheck = new ArrayList<>();
+
+        private Integer asyncReportDelay;
+
+        public Builder withTextDataCheck(SandboxDocumentTextDataCheck textDataCheckReport) {
```
davidgrayston, line 88: Should we update the method too?

```suggestion
        public Builder withDocumentTextDataCheck(SandboxDocumentTextDataCheck textDataCheckReport) {
```

yoti-sdk-sandbox/src/main/java/com/yoti/api/client/sandbox/docs/request/SandboxTaskResults.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Builder for {@link SandboxTaskResults}
+     */
+    public static class Builder {
+
+        private List<SandboxDocumentTextDataExtractionTask> textDataExtractionTasks = new ArrayList<>();
+
+        public Builder withTextDataExtractionTask(SandboxDocumentTextDataExtractionTask textDataExtractionTask) {
```
davidgrayston, line 37:
```suggestion
        public Builder withDocumentTextDataExtractionTask(SandboxDocumentTextDataExtractionTask textDataExtractionTask) {
```

yoti-sdk-sandbox/src/main/java/com/yoti/api/client/sandbox/docs/request/check/SandboxCheckBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+
+    protected SandboxRecommendation recommendation;
+    protected List<SandboxBreakdown> breakdown;
+
+    public T withRecommendation(SandboxRecommendation recommendation) {
+        this.recommendation = recommendation;
+        return self();
+    }
+
+    public T withBreakdown(SandboxBreakdown breakdown) {
```
echarrod, line 19: I can do this in .NET for the "with" methods in this class:
```cs
public SandboxCheckBuilder<TCheck> WithBreakdown(SandboxBreakdown breakdown)
        {
            Breakdown.Add(breakdown);
            return this;
        }
```
Presume you can't do the same in Java to avoid having to use generics?

echarrod: (...although I do have a generic on the build() method)

yoti-sdk-sandbox/src/main/java/com/yoti/api/client/sandbox/docs/request/check/SandboxCheckBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+        if (this.breakdown == null) {
+            this.breakdown = new ArrayList<>();
+        }
```
davidgrayston, line 22: @MrBurtyyy as discussed, we can default this to an empty list

@echarrod .NET validates this is not an empty list, which might not be necessary (if we feel the SDK should validate this, we need to update PHP/Node)

yoti-sdk-sandbox/src/main/java/com/yoti/api/client/sandbox/docs/request/check/SandboxCheckBuilder.java
MrBurtyyy: We don't have the generic as the check, rather the builder.  We need to know the type of the builder so we can return the correct child class from `with` methods (see `self()` abstract).

echarrod: Ah, I did actually think this was fine in .NET, but I've since realised that once you've used a parent `With`, you can't use the child `With`, so it's possible, but you'd have to order them. So will change that in .NET now

MrBurtyyy: Yep, that's why we have the builder as the generic and not the check itself 👍

PR #211, opened 2020-09-04
SDK-1743: Add support for blocking the collection of bio-metric consent

MrBurtyyy:


yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/session/create/SimpleSessionSpec.java
```diff
@@ -35,14 +36,18 @@
     @JsonProperty("required_documents")
     private final List<RequiredDocument> requiredDocuments;
 
+    @JsonProperty("block_biometric_consent")
+    private final Boolean blockBiometricConsent;
```
davidgrayston, line 40: Just an observation: after some discussions for the Go SDK, the conclusion was to default newly introduced `boolean` values to `false`, which might mean you could use the primitive type here?

The way you have it means it's treated as optional (omitted from payload), which I think is more consistent with the rest of the Java SDK, so I think it's fine to leave this as is.

davidgrayston (approved):

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/session/create/SimpleSessionSpec.java
bucky-boy: None of the SDKs should be providing their own default values.  Default values are decided by our _session-svc_, not the SDKs.  If the _Relying Business_ have not explicitly specified a value then the SDK should not be sending anything.

It's particularly important that this behaviour is observed for the this new _biometric_consent_ flag, as well as _manual_check_ on the authenticity check config.

bucky-boy (approved):
This looks ok to me, but both this and the _manual_check_ flag need to be tested properly before this can be released.  Neither should be choosing their own default value.

yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/session/create/SimpleSessionSpec.java
echarrod: Let's leave the booleans as non-primitive types, as they are currently, to follow how the API was designed. They then can be excluded for Doc Scan API calls. We can discuss the profile service default values separately

echarrod (approved):

PR #218, opened 2020-09-21
IN-3119: java Example QR

nikhilPank:


echarrod (approved):

davidgrayston (changes requested):
Do we need to make this change in the dynamic share example too?

https://github.com/getyoti/yoti-java-sdk/blob/DEVELOPMENT/yoti-sdk-spring-boot-example/src/main/resources/templates/dynamic-share.html#L73

davidgrayston (approved):

PR #221, opened 2020-10-05
SDK-1784: Add sandbox support for NFC

MrBurtyyy:


yoti-sdk-sandbox/src/main/java/com/yoti/api/client/sandbox/docs/request/task/SandboxDocumentTextDataExtractionTask.java
```diff
@@ -67,10 +69,15 @@ public Builder withDocumentFilter(SandboxDocumentFilter documentFilter) {
             return this;
         }
 
+        public Builder withDocumentIdPhoto(String contentType, byte[] documentIdPhoto) {
+            String b64DocumentIdPhoto = Base64.getEncoder().encodeToString(documentIdPhoto);
```
davidgrayston, line 73: Should base64 encoding the image be the responsibility of `SandboxDocumentIdPhoto()`?

MrBurtyyy: I think I would prefer the built object to just be a container for the information, rather than having to know it needs to Base64 encode the bytes.  The constructor of the bean has package level visibility, so I don't see this being a problem where a user tries to use the class directly (they should be using the builder).

davidgrayston (approved):

echarrod (approved):

PR #222, opened 2020-10-05
SDK-1774: Allow retrieval of frame for an id document page

MrBurtyyy:


davidgrayston (approved):

echarrod (approved):

davidgrayston (changes requested):
We also need to update the demo to display the frames

davidgrayston (approved):
@MrBurtyyy I've copied the HTML from other SDK demos in 42de1a6 for multiple pages with frames

echarrod (approved):

PR #234, opened 2020-12-07
SDK-1893: Add auth type to notification config

MrBurtyyy:


bucky-boy (approved):
Looks good to me, approved.

I presume you've checked that it's not going to send any _auth_type_ if nothing has been set...?

PR #240, opened 2021-01-05
SDK-1855: Add Github Actions file for Junit tests

MrBurtyyy:


echarrod (approved):
Nice :ok_hand: 

@emmas-yoti we should make sure we update the badges too, like Alex has done here.

Also we'll need to update any references to Travis in the README. (I think we mention it when talking about version support in some of them)

PR #242, opened 2021-01-05
SDK-1855: Add sonar scanner action, and change branches actions will execute against

MrBurtyyy:


bucky-boy (approved):

echarrod (approved):
Have removed the requirement for Java 7 checks to pass before merging, since we're dropping support. Doesn't look like it's propagated just yet, but should be soon

PR #251, opened 2021-01-13
SDK-1861: Add static methods to abstract classes as syntactical sugar

MrBurtyyy:


MrBurtyyy:
Closing as we don't quite think this is the right way of handling this.  Builders are still available through their respective child classes.

PR #258, opened 2021-01-18
SDK-1370: Ensure SDK handles RFC3339 format when parsing expiry date

MrBurtyyy:


echarrod (approved):

yoti-sdk-api/src/main/java/com/yoti/api/client/DateTime.java
```diff
@@ -32,6 +36,17 @@ public static DateTime from(long microseconds) {
         return new DateTime(Date.from(calendar), Time.from(calendar, mod));
     }
 
+    public static DateTime from(String dateTimeStringValue) throws DateTimeParseException {
```
bucky-boy, line 39: I'm not sure this public method should be returning null if gets an invalid string (i.e. null or empty).

What's the background to this change?  What's the context?

bucky-boy (approved):

PR #269, opened 2021-03-25
SDK-1897: Add WATCHLIST_SCREENING check

MrBurtyyy:


yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/retrieve/WatchlistScreeningSummaryResponse.java
```diff
@@ -0,0 +1,10 @@
+package com.yoti.api.client.docs.session.retrieve;
+
+public class WatchlistScreeningSummaryResponse extends WatchlistSummaryResponse<WatchlistScreeningSearchConfigResponse> {
+
+    @Override
+    public WatchlistScreeningSearchConfigResponse getSearchConfig() {
```
bucky-boy, line 6: Why do we need this, if the base class already returns it?

MrBurtyyy: The base class returns a generic `Config extends WatchlistSearchConfigResponse` which is an abstract class with default package scope (meaning the _relying business_ can't use it).  We should consider making it public, or use this method

bucky-boy (approved):

PR #271, opened 2021-04-26
SDK-1912: Add support for WATCHLIST_ADVANCED_CA check

MrBurtyyy:


yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/check/advanced/RequestedTypeListSources.java
```diff
@@ -0,0 +1,35 @@
...
+
+    RequestedTypeListSources(List<String> types) {
+        this.types = types;
+    }
+
+    public List<String> getTypes() {
+        return types;
+    }
+
+    public static class Builder {
```
bucky-boy, line 20: Do we want a _builder()_ method on this class too?

And the others that extend some abstract base class with the builders on it

bucky-boy (approved):

PR #303, opened 2022-03-23
SDK-2077 SDK-2086: Add identity profile to dynamic policy builder && subject to dynamic scenario builder

irotech:


yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/DynamicPolicy.java
```diff
@@ -10,34 +10,40 @@
...
-    @JsonProperty("wanted_remember_me")
+    @JsonProperty(Property.WANTED_REMEMBER_ME)
     private final boolean wantedRememberMe;
 
-    @JsonProperty("wanted_remember_me_optional")
+    @JsonProperty(Property.WANTED_REMEMBER_ME_OPTIONAL)
     private final boolean wantedRememberMeOptional;
 
+    @JsonProperty(Property.IDENTITY_PROFILE_REQUIREMENTS)
+    private final IdentityProfile identityProfile;
```
BenSmithYoti, line 35: If the wantedRememberMe field and constructor param are best with "Optional" suffix - should we perhaps similarly mark the new identityProfile as such with "identityProfileOptional"?

yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/profile/IdentityProfile.java
```diff
@@ -0,0 +1,76 @@
+package com.yoti.api.client.shareurl.policy.profile;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Set of data required to represent an identity profile
+ */
+public final class IdentityProfile {
```
BenSmithYoti, line 8: In dotnet we have modelled this intentionally "loosely" using object (as opposed to having a class like IdentityProfile definied for the input). https://github.com/getyoti/yoti-dotnet-sdk/pull/354/commits/cc94506bc9496c878c8fbdfffddb464eb91b4a17 This means  Integrators can supply an anonymous object to match their needs against the API's schema; i.e. the SDK is buffered from potential changes to the json schema for the request. I am happy to approve as is, only hesitation is that the java SDK might need to be updated if these inputs change. @echarrod thoughts here please?

yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/DynamicPolicy.java
```diff
@@ -10,34 +10,40 @@
...
+
     DynamicPolicy(Collection<WantedAttribute> wantedAttributes,
             Set<Integer> wantedAuthTypes,
             boolean wantedRememberMe,
-            boolean wantedRememberMeOptional) {
+            boolean wantedRememberMeOptional,
+            IdentityProfile identityProfile) {
```
BenSmithYoti, line 41: I notice no explicit "public" here on the constructor - so is this constructor visible to code from other packages? The new constuctor parameter is intentionally optional (with a default of null) in dotnet (where the constructor is public and is visible to consuming code). If this constuctor is visible to integrators then perhaps we should implement something similar here in Java - using overload(s) for the optional params? Certainly on dotnet side we have various Integrators using various classes directly (without their builders). If this is possible on the Java side shall we avoid breaking changes for such Integrators? Conversely, if this constructor is not visible to integrators code (and they must use the Builder to construct) then I don't see the need to change anything here.

BenSmithYoti: Appologies I am not yet fully setup with my Java IDE and yoti-java-sdk otherwise I would answer this question about visibility myself.

yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/profile/IdentityProfile.java
```diff
@@ -0,0 +1,76 @@
...
+    public static final class Builder {
+
+        private String framework;
+        private IdentityProfileScheme scheme;
+
+        private Builder() { }
+
+        public Builder withFramework(String framework) {
+            this.framework = framework;
+
```
BenSmithYoti, line 52: Minor: I notice some cases here we have a line break before the 
`return this;`
...and some we do not. Is it worth standardising quickly?

yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/DynamicPolicy.java
```diff
@@ -260,11 +276,34 @@ public Builder withWantedRememberMeOptional(boolean wantedRememberMeOptional) {
...
+                    wantedAuthTypes,
+                    wantedRememberMe,
+                    wantedRememberMeOptional,
+                    identityProfile
+            );
         }
 
     }
 
+    private static final class Property {
```
BenSmithYoti, line 297: I generally like this (pushing magic strings and magic numbers down to a single source of truth / constant / static final etc), however, I do not remember seeing the same pattern being used in the java SDK for these @JsonProperty("X") declarations. I believe each field here is only being used once (so we are not making anything DRY-er), and seeing as the only difference between the field names and the string values is the casing - I wonder if for consistency and simplicity we would not be better off in this case simply using the relevant strings for each JsonProperty() directly?

BenSmithYoti (commented):
Looks good - thanks Vince. Have added a few questions and suggestions. Thanks.

yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/DynamicPolicy.java
echarrod: @BenSmithYoti I _think_ this is in fact saying that "the wanted remember me is optional in the share", rather than _itself_ being an optional property which may or not be provided for the policy to be built

yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/profile/IdentityProfile.java
echarrod: Yes I believe @pgreci said this should be dynamic, as the structure might change in future iterations, and we don't want to have to update the SDKs as we add new properties to this object :+1:

BenSmithYoti: So I am not sure about the best way to achieve this dynamic nature in Java. In dotnet we simply used object for the type so Integrators can then construct with anonymus syntax as per CreateStandardIdentityProfileRequirements here https://github.com/getyoti/yoti-dotnet-sdk/pull/354/commits/cc94506bc9496c878c8fbdfffddb464eb91b4a17 @irotech what are your thoughts on what is best in Java here?

yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/DynamicPolicy.java
BenSmithYoti: I see this pattern in use elsewhere in the Java SDK code already, so marking this as resolved.

yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/DynamicPolicy.java
irotech: > If this is possible on the Java side shall we avoid breaking changes for such Integrators?
I wouldn't say it's a breaking change when it comes to different languages. Not an expert in out SDKs but I don't think you can actually reuse anything from .NET to Java and viceversa

As general concept providing a Builder and restricting the constructor to package/private is a good practice to focus the validation in both Builder.setters (validity of the data provided) and Builder.build (presence of mandatory properties).

yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/DynamicPolicy.java
irotech: This is more a team's mindset/style. In this specific scenario is not really required as it is used only once but usually we De/Serialize JSONs using the same POJO.

BenSmithYoti: Sure - makes perfect sense.

yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/DynamicPolicy.java
pn-santos: > I don't know if this makes said constructor "private to the package" or visible to an Integrators consuming code?)

It does.

BenSmithYoti: Ok perfect. Thanks

irotech:
Rebased and changed the target branch to DEVELOPMENT

yoti-sdk-parent/pom.xml
```diff
@@ -101,7 +100,7 @@
     <!-- maven + plugin versions -->
     <maven.compiler.version>3.8.1</maven.compiler.version>
     <maven.source.version>2.2.1</maven.source.version>
```
pn-santos, line 102: these also have more recent versions, was there a reason they weren't upgraded too?

yoti-sdk-spring-security/src/test/java/com/yoti/api/spring/security/web/YotiAuthenticationFilterTest.java
```diff
@@ -28,6 +24,9 @@
 
     private YotiAuthenticationFilter filter;
 
+    @Rule
+    public MockitoRule rule = MockitoJUnit.rule();
```
pn-santos, line 28: What's the reason behind this change? What was wrong with `@RunWith` (seems to be the convention used by all other UTs -- at leas the random ones I checked)?

I would suggest to stick with the conventions. A change like this (where the same pattern is used elsewhere) should be done everywhere or not at all. Otherwise as times goes one the codebase becomes a myriad of different solutions to the same problem for no reason at all.

yoti-sdk-spring-security/src/test/java/com/yoti/api/spring/security/web/YotiAuthenticationFilterTest.java
```diff
@@ -79,6 +78,7 @@ private YotiAuthenticationToken doAuthenticationAndCaptureToken() {
 
     private void initMocks() {
         when(mockRequest.getParameter(TOKEN_PARAM)).thenReturn(TOKEN);
-        when(mockDetailsSource.buildDetails(any(HttpServletRequest.class))).thenReturn(null);
+        when(mockDetailsSource.buildDetails(Mockito.any(HttpServletRequest.class))).thenReturn(null);
```
pn-santos, line 81: Mockito methods should be statically imported. This change should be reverted.

yoti-sdk-parent/pom.xml
```diff
@@ -81,15 +81,14 @@
     <!-- sdk dependency versions -->
     <slf4j.version>1.7.30</slf4j.version>
```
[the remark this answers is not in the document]
irotech: https://yoti.slack.com/archives/G018LQMJ0D7/p1648220340152019

I had to upgrade few due CVE errors in packaging. As per slack comment I could upgrade all the dependencies and the plug-ins in a separate commit.

yoti-sdk-parent/pom.xml
```diff
@@ -81,15 +81,14 @@
...
     <bouncy.castle.version>1.60</bouncy.castle.version>
-    <jackson.version>2.9.10.7</jackson.version>
-    <jackson-jsr310.version>2.9.10</jackson-jsr310.version>
-    <google.protobuf.version>3.14.0</google.protobuf.version>
+    <jackson.version>2.13.0</jackson.version>
```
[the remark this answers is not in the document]
irotech: latest core is 2.13.2 --> (https://github.com/FasterXML/jackson-core)
latest databind 2.13.2.1 --> (https://github.com/FasterXML/jackson-databind)

yoti-sdk-parent/pom.xml
```diff
@@ -101,7 +100,7 @@
     <!-- maven + plugin versions -->
     <maven.compiler.version>3.8.1</maven.compiler.version>
     <maven.source.version>2.2.1</maven.source.version>
-    <maven.enforcer.version>3.0.0-M1</maven.enforcer.version>
+    <maven.enforcer.version>3.0.0</maven.enforcer.version>
```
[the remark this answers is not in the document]
irotech: it seems to work fine with 
maven.enforcer.version -->3.0.0 
mojo.extra.rules.version -->1.5.1

pn-santos (approved):

echarrod (approved):

PR #304, opened 2022-03-24
SDK-2086: Add subject to dynamic policy builder

irotech:


yoti-sdk-api/src/test/java/com/yoti/api/client/shareurl/policy/DynamicPolicyTest.java
```diff
@@ -306,6 +307,17 @@ public void buildWithIdentityProfile() {
         assertThat(result.getIdentityProfile().getScheme(), is(equalTo(identityProfileScheme)));
     }
 
+    @Test
+    public void buildWithSubject() {
```
laurent-yoti, line 311: @irotech I am not really sure to understand why is the subject part of the policy, as it seems to be living in the scenario instead (see [dot-net for example](https://github.com/getyoti/yoti-dotnet-sdk/pull/354/files))

yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/DynamicPolicy.java
```diff
@@ -34,16 +35,21 @@
...
 
+    @JsonProperty(Property.SUBJECT)
+    private final Subject subject;
+
     DynamicPolicy(Collection<WantedAttribute> wantedAttributes,
             Set<Integer> wantedAuthTypes,
             boolean wantedRememberMe,
             boolean wantedRememberMeOptional,
-            IdentityProfile identityProfile) {
+            IdentityProfile identityProfile,
```
laurent-yoti, line 45: Shouldn't we stick with `IdentityProfileRequirements` ?

PR #306, opened 2022-03-29
SDK-2092: Add attribute by id retrieval

irotech:


irotech:
Looking into the code smells

yoti-sdk-api/src/test/java/com/yoti/api/client/HumanProfileTest.java
```diff
@@ -374,6 +376,43 @@ public void getIdentityProfileReport_shouldReturnValueForIdentityProfileReport()
...
+
+        Attribute<?> attribute = humanProfile.getAttributeById("unknown_id");
+
+        assertNull(attribute);
+    }
+
+    private static Attribute<JpegAttributeValue> createImageAttribute(String id) {
+        return new Attribute<>(
+                id,
+                String.format("name_%s", id),
```
echarrod, line 411: Could add a test for adding two attributes with the same name? This is what will be returned on the profile, so we will need to test for that 👍

irotech: sure

PR #309, opened 2022-03-31
SDK-2107: Add IDV response identity profile retriever

irotech:


pn-santos (approved):
LGTM but I'm not really familiar with the IDV stuff

BenSmithYoti (approved):
LGTM. Only thing to mention -which I'm sure you are aware of already - is that you have implemented classes for the responses. In dotnet we have keep things loose. That's absolutely fine --as long as someone updates the Java SDK if the form of the responses change.

irotech:
> LGTM. Only thing to mention -which I'm sure you are aware of already - is that you have implemented classes for the responses. In dotnet we have keep things loose. That's absolutely fine --as long as someone updates the Java SDK if the form of the responses change.

I actually asked in advance and took the .NET as reference so that we have GetSessionResult.IdentityProfileResponse.identityProfileReport loose

```
GetSessionResult
	IdentityProfileResponse identityProfile
		String subjectId
		String result
		IdentityProfileFailureResponse failureReason
			String reasonCode
		Object identityProfileReport <-- <any JSON Structure>

echarrod:
> LGTM. Only thing to mention -which I'm sure you are aware of already - is that you have implemented classes for the responses. In dotnet we have keep things loose. That's absolutely fine --as long as someone updates the Java SDK if the form of the responses change.

Yep, I checked with Paolo and it's the "Identity Profile Report" that we want to keep as JSON, the other elements are more fixed and we were happy with adding those as properties :+1:

PR #320, opened 2022-04-25
SDK-2115: Support 3rd party Identity fraud check result

irotech:


yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/check/RequestedThirdPartyIdentityFraudOneCheck.java
```diff
@@ -0,0 +1,41 @@
+package com.yoti.api.client.docs.session.create.check;
+
+import com.yoti.api.client.docs.DocScanConstants;
+
+public class RequestedThirdPartyIdentityFraudOneCheck extends RequestedCheck<RequestedThirdPartyIdentityFraudOneConfig> {
```
MrBurtyyy, line 5: `THIRD_PARTY_IDENTITY_FRAUD_1` checks cannot be requested like regular checks (they are part of the Identity Profile flow).  This isn't needed :)

MrBurtyyy (changes requested):
This check cannot be requested like traditional checks, so we can remove any bits to do with requesting the check

yoti-sdk-api/src/test/java/com/yoti/api/client/docs/session/create/check/CheckType.java
```diff
@@ -0,0 +1,13 @@
+package com.yoti.api.client.docs.session.create.check;
+
+public enum CheckType {
```
BenSmithYoti, line 3: Think this is unused now that the extra test file ..api/client/docs/session/create/check/RequestedThirdPartyIdentityFraudOneCheckTest.java has been removed with second commit

BenSmithYoti (changes requested):
Comment on now (I believe) unused file. 
Other than that LGTM. Thanks Vince.

BenSmithYoti:
Oh, just noticed (waiting to see what @echarrod says for dotnet side) but we have this HTML template file for the IDV demo app https://github.com/getyoti/yoti-dotnet-sdk/blob/master/src/Examples/DocScan/DocScanExample/Views/IdVerify/Success.cshtml which I feel might need an update to include this new type of check. If we do want that I believe that there is an equivalent (index.html maybe?) in the java repo. @echarrod any input on this would be great please, then I can finish my PR. Thanks

BenSmithYoti:
@MrBurtyyy do you know about this perhaps? Is this demo UI wanting to be updated in this iteration? https://github.com/getyoti/yoti-java-sdk/pull/320#issuecomment-1109708226

echarrod:
> Oh, just noticed (waiting to see what @echarrod says for dotnet side) but we have this HTML template file for the IDV demo app https://github.com/getyoti/yoti-dotnet-sdk/blob/master/src/Examples/DocScan/DocScanExample/Views/IdVerify/Success.cshtml which I feel might need an update to include this new type of check. If we do want that I believe that there is an equivalent (index.html maybe?) in the java repo. @echarrod any input on this would be great please, then I can finish my PR. Thanks

Yep we'll need that so Darren to verify that this new check comes through in the response correctly :+1:

BenSmithYoti (approved):
LGTM

PR #322, opened 2022-04-28
SDK-2135: Add DBS check example page

irotech:
Example, available at https://localhost:8443/dbs-check , that builds a standard DBS check via dynamic scenario builder
```
{
    "scheme": {
        "objective": "STANDARD",
        "type": "DBS"
    },
    "trust_framework": "UK_TFIDA"
}
```

### DIFF
--- a/yoti-sdk-spring-boot-example/src/main/java/com/yoti/api/examples/springboot/YotiLoginController.java
+++ b/yoti-sdk-spring-boot-example/src/main/java/com/yoti/api/examples/springboot/YotiLoginController.java
@@ -1,6 +1,10 @@
 package com.yoti.api.examples.springboot;
 
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.stream.Collectors;
 
 import com.yoti.api.attributes.AttributeConstants;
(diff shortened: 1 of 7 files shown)


yoti-sdk-spring-boot-example/src/main/java/com/yoti/api/examples/springboot/YotiLoginController.java
```diff
@@ -152,6 +158,43 @@ public String doLogin(@RequestParam("token") final String token, final Model mod
         return "profile";
     }
 
+    @RequestMapping("/dbs-standard")
```
echarrod, line 161: I updated the .NET example (in [this commit](https://github.com/getyoti/yoti-dotnet-sdk/pull/375/commits/e8092a1664dbe0f0c858263bf7315de4cfbbc61a)) to be `dbs-check` rather than `dbs-standard`, as I think it makes more sense from a client's perspective, they're just wanting to see a DBS check in action:
```suggestion
    @RequestMapping("/dbs-check")
```

echarrod (approved):
Looks good Vince. Just one comment, where it would be good to be consistent across SDKs :+1:

yoti-sdk-spring-boot-example/src/main/java/com/yoti/api/examples/springboot/YotiLoginController.java
irotech: oh I see, sure no problem

PR #327, opened 2022-05-10
SDK-2141: Add error code and description in failure receipt message

irotech:
Trying to test a failure receipt. Gonna add more test coverage to make Sonar happy.

### DIFF
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/ReceiptFetcher.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/ReceiptFetcher.java
@@ -5,9 +5,11 @@ import static com.yoti.api.client.spi.remote.call.YotiConstants.DEFAULT_CHARSET;
 import java.security.KeyPair;
 import java.security.PrivateKey;
 import java.util.Base64;
+import java.util.Optional;
 
 import com.yoti.api.client.ActivityFailureException;
 import com.yoti.api.client.ProfileException;
+import com.yoti.api.client.spi.remote.call.ProfileResponse;
 import com.yoti.api.client.spi.remote.call.ProfileService;
 import com.yoti.api.client.spi.remote.call.Receipt;
 import com.yoti.api.client.spi.remote.util.DecryptionHelper;
@@ -31,11 +33,13 @@ public class ReceiptFetcher {
 
     public Receipt fetch(String encryptedConnectToken, KeyPair keyPair, String appId) throws ProfileException {
         LOG.debug("Decrypting connect token: '{}'", encryptedConnectToken);
+
         String connectToken = decryptConnectToken(encryptedConnectToken, keyPair.getPrivate());
         LOG.debug("Connect token decrypted: '{}'", connectToken);
-        Receipt receipt = profileService.getReceipt(keyPair, appId, connectToken);
-        validateReceipt(receipt, connectToken);
-        return receipt;
+
+        ProfileResponse profile = profileService.getProfile(keyPair, appId, connectToken);
+        validateReceipt(profile, connectToken);
+        return profile.getReceipt();
     }
 
     private String decryptConnectToken(String encryptedConnectToken, PrivateKey privateKey) throws ProfileException {
@@ -43,17 +47,23 @@ public class ReceiptFetcher {
             byte[] byteValue = Base64.getUrlDecoder().decode(encryptedConnectToken);
             byte[] decryptedToken = DecryptionHelper.decryptAsymmetric(byteValue, privateKey);
             return new String(decryptedToken, DEFAULT_CHARSET);
-        } catch (Exception e) {
-            throw new ProfileException("Cannot decrypt connect token", e);
+        } catch (Exception ex) {
+            throw new ProfileException("Cannot decrypt connect token", ex);
         }
     }
 
-    private void validateReceipt(Receipt receipt, String connectToken) throws ProfileException {
-        if (receipt == null) {
-            throw new ProfileException("No receipt for '" + connectToken + "' was found");
-        }
-        if (receipt.getOutcome() == null || !receipt.getOutcome().isSuccessful()) {
-            throw new ActivityFailureException("Sharing activity unsuccessful for " + receipt.getDisplayReceiptId());
+    private void validateReceipt(ProfileResponse profile, String connectToken) throws ProfileException {
+        Receipt receipt = Optional.ofNullable(profile)
+                .map(ProfileResponse::getReceipt)
+                .orElseThrow(() -> new ProfileException("No profile for '" + connectToken + "' was found"));
+
+        if (!receipt.hasOutcome(Receipt.Outcome.SUCCESS)) {
+            throw new ActivityFailureException(
+                    String.format("Sharing activity unsuccessful for %s%s",
+                            receipt.getDisplayReceiptId(),
+                            Optional.ofNullable(profile.getError()).map(e -> String.format(" - %s", e)).orElse("")
+                    )
+            );
         }
     }
 
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ErrorDetails.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ErrorDetails.java
@@ -1,0 +1,88 @@
+package com.yoti.api.client.spi.remote.call;
+
+import java.util.Objects;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+@JsonDeserialize(builder = ErrorDetails.Builder.class)
+public final class ErrorDetails {
+
+    private final String code;
+    private final String description;
+
+    private ErrorDetails(Builder builder) {
+        this.code = builder.code;
+        this.description = builder.description;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public String getCode() {
+        return code;
+    }
+
+    public String getDescription() {
+        return description;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        ErrorDetails that = (ErrorDetails) o;
+        return Objects.equals(code, that.code) && Objects.equals(description, that.description);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(code, description);
+    }
+
+    @Override
+    public String toString() {
+        return String.format("Error[code='%s', description='%s']", code, description);
+    }
+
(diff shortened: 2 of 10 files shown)


yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/ReceiptFetcher.java
```diff
@@ -31,29 +33,40 @@ public static ReceiptFetcher newInstance() {
...
-            throw new ActivityFailureException("Sharing activity unsuccessful for " + receipt.getDisplayReceiptId());
+
+        Receipt receipt = profile.getReceipt();
+
+        if (!receipt.hasSuccessOutcome()) {
+            ErrorDetails error = profile.getError();
+            throw new ActivityFailureException(
+                    error == null
+                            ? String.format("Sharing activity unsuccessful for %s", receipt.getDisplayReceiptId())
+                            : String.format("Sharing activity unsuccessful for %s  %s", receipt.getDisplayReceiptId(), error)
```
pn-santos: Could be `DRY`er:
```java
String.format(
        "Sharing activity unsuccessful for %s%s", 
        receipt.getDisplayReceiptId(), 
        Optional.ofNullable(error).map(ErrorDetaisL::toString).orElse("")
)
```

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/ReceiptFetcher.java
```diff
@@ -31,29 +33,40 @@ public static ReceiptFetcher newInstance() {
...
-        if (receipt == null) {
-            throw new ProfileException("No receipt for '" + connectToken + "' was found");
+    private void validateReceipt(ProfileResponse profile, String connectToken) throws ProfileException {
+        if (profile == null || profile.getReceipt() == null) {
+            throw new ProfileException("No profile for '" + connectToken + "' was found");
         }
-        if (receipt.getOutcome() == null || !receipt.getOutcome().isSuccessful()) {
-            throw new ActivityFailureException("Sharing activity unsuccessful for " + receipt.getDisplayReceiptId());
+
+        Receipt receipt = profile.getReceipt();
```
pn-santos, line 60:
```java
Receipt receipt = Optional.ofNullable(profile)
        ,map(ProfileResponse::getReceipt)
        .orElseThrow(() -> new ProfileException(...));

File: yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ErrorDetails.java [hunk 3]
```diff
@@ -0,0 +1,94 @@
...
+
+public final class ErrorDetails {
+
+    @JsonProperty(Property.ERROR_CODE)
+    private String code;
+
+    @JsonProperty(Property.DESCRIPTION)
+    private String description;
+
+    private ErrorDetails() { }
```
pn-santos, line 15: why the need for both the builder and the no arg ctor?

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ProfileResponse.java
```diff
@@ -33,28 +37,46 @@ public void setReceipt(Receipt receipt) {
...
-        if (!sessionData.equals(that.sessionData)) return false;
-        return receipt.equals(that.receipt);
-
+        if (!sessionData.equals((that.sessionData))) {
+            return false;
+        }
+        if (!receipt.equals((that.receipt))) {
+            return false;
+        }
+        return error.equals(that.error);
```
pn-santos, line 64: Never return a boolean as a result of a if condition!
```java
return sessionData.equals(that.sessionData) 
        && receipt.equals(that.receipt) 
        && error.equals(that.error);
```

Also AFAICT there's no assurance that any of these properties are not `null` so the comparison should be done using `Objects.equals`, e.g. `Objects.equals(sessionData, that.sessionData)`

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ProfileResponse.java
```diff
@@ -33,28 +37,46 @@ public void setReceipt(Receipt receipt) {
...
     }
 
     @Override
     public int hashCode() {
         int result = sessionData.hashCode();
         result = 31 * result + receipt.hashCode();
+        result = 31 * result + error.hashCode();
         return result;
```
pn-santos, line 72:
```java
return Objects.hash(sessionData, receipt, error);
```
(applies to all `hashCode` implementations)

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ProfileResponse.java
```diff
@@ -2,19 +2,23 @@
...
+    private ErrorDetails error;
 
-    private ProfileResponse(String sessionData, Receipt receipt){
+    public ProfileResponse() { }
+
+    private ProfileResponse(String sessionData, Receipt receipt, ErrorDetails error) {
         this.sessionData = sessionData;
         this.receipt = receipt;
+        this.error = error;
     }
```
pn-santos, line 22: 1. Why have the empty `public` ctor?
2. Why not pass the builder instance to the private ctor?

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ErrorDetails.java
```diff
@@ -0,0 +1,94 @@
...
+
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return String.format("- Error[code='%s', description='%s']", code, description);
+    }
+
+    public static final class Builder {
```
pn-santos, line 62: Shouldn't this builder be used to de-serialize a JSON object? The class isn't specifying it with `@JsonDeserialize`

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ErrorDetails.java
```diff
@@ -0,0 +1,94 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public final class ErrorDetails {
+
+    @JsonProperty(Property.ERROR_CODE)
+    private String code;
+
+    @JsonProperty(Property.DESCRIPTION)
+    private String description;
```
pn-santos, line 13: Isn't this class supposed to _de_-serialize a JSON? If so the `JsonProperties` should go in the builder methods (assuming the builder is intended to be used to de-serialize the JSON)

If this class is intended to serialize to JSON then nevermind (if both then the comment regarding the builder being used for de-serialization still applies).

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/Receipt.java
```diff
@@ -7,31 +7,36 @@
...
     private byte[] rememberMeId;
-    @JsonProperty("parent_remember_me_id")
+    @JsonProperty(Property.PARENT_REMEMBER_ME_ID)
     private byte[] parentRememberMeId;
-    @JsonProperty("sharing_outcome")
+    @JsonProperty(Property.SHARING_OUTCOME)
     private Outcome outcome;
-    @JsonProperty("timestamp")
+    @JsonProperty(Property.TIMESTAMP)
     private String timestamp;
```
pn-santos, line 34: (nit) maybe also add a newline before every `@JsonProperty`

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/Receipt.java
```diff
@@ -7,31 +7,36 @@
...
 
+    public boolean hasSuccessOutcome() {
+        return Outcome.SUCCESS == outcome;
+    }
```
pn-santos, line 38: Would be better has:
```java
public boolean hasOutcome(Outcome outcome) {
    return this.outcome == outcome;
}
```
for a couple of reasons:
- Even if now there's only two outcomes in the future could be more 
- The code that calls this method becomes more readable (and flexible) if it's able to check for either success or failure (or whatever other outcomes may exist in the future)
- Hardcoding state (the value of an enum) into a method name is best avoided as it creates unnecessary coupling between the code and state.

yoti-sdk-api/src/test/java/com/yoti/api/client/spi/remote/call/ReceiptOutcome.java
```diff
@@ -0,0 +1,9 @@
+package com.yoti.api.client.spi.remote.call;
+
+public enum ReceiptOutcome {
+    SUCCESS(true),
+    FAILURE(false);
+
+    ReceiptOutcome(boolean successful) { }
```
pn-santos, line 7: what's the point of having the vars if they won't be set or accessed? Why not just:
```java
public enum ReceiptOutcome {
    
    SUCCESS,
    FAILURE;

}
```

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ErrorDetails.java
```diff
@@ -39,24 +36,19 @@ public boolean equals(Object o) {
...
-        result = prime * result + ((description == null) ? 0 : description.hashCode());
-
-        return result;
+        return Objects.hash(code, description);
     }
 
     @Override
     public String toString() {
-        return String.format("- Error[code='%s', description='%s']", code, description);
+        return String.format(" - Error[code='%s', description='%s']", code, description);
```
pn-santos, line 51: (nit) Why does this have a ` - ` at the start of the string? I'd expect a `toString` to no have any "prefix" (callers can always prefix it in the way they see fit)

---

(nit) Also probably easier to survive refactorings if this was 
```java
%s[code='%s' ...]", this.getClass().getSimpleName(), code, ...)
```

pn-santos (approved):

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ErrorDetails.java
irotech: gonna change as suggested and handle spaces or `-` in the caller. I just managed to test and the final result is 

image

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ErrorDetails.java
```diff
@@ -0,0 +1,88 @@
...
+    private ErrorDetails(Builder builder) {
+        this.code = builder.code;
+        this.description = builder.description;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public String getCode() {
```
echarrod, line 23: Will this always be an integer? Was just thinking it would be easier for clients to switch on, if it's an int

echarrod (approved):

PR #357, opened 2023-02-01
Dependencies version bump

irotech:


yoti-sdk-parent/pom.xml
```diff
@@ -79,22 +79,24 @@
     <maven.compiler.target>${supported.java.version}</maven.compiler.target>
     
     <!-- sdk dependency versions -->
-    <slf4j.version>1.7.36</slf4j.version>
+    <slf4j.version>2.0.6</slf4j.version>
```
pn-santos, line 82: FYI: This makes Java 8 the min required version for the JDK. I'm fine with it but just pointing this out in case supporting Java 7 is needed (if so it needs to be 1.7.36)

yoti-sdk-parent/pom.xml
```diff
@@ -79,22 +79,24 @@
...
     <!-- spring versions -->
-    <spring-framework.boot.version>2.7.1</spring-framework.boot.version>
-    <spring-framework.version>5.3.21</spring-framework.version>
-    <spring-framework.security.version>5.7.2</spring-framework.security.version>
+    <!-- Spring Framework 6.0.x requires JDK 17+   -->
+    <spring-framework.version>5.3.25</spring-framework.version>
+    <!-- Spring Framework Boot 3.0.x requires JDK 17+   -->
+    <spring-framework.boot.version>2.7.8</spring-framework.boot.version>
+    <!-- Spring Framework Security 6.0.x requires JDK 17+   -->
+    <spring-framework.security.version>5.8.1</spring-framework.security.version>
```
pn-santos, line 94: Best to use springboot bom to define spring-related versions imo:
```xml
<dependencyManagement>
  <dependencies>
    <dependency>
      <groupId>org.springframework.boot</groupId>
      <artifactId>spring-boot-dependencies</artifactId>
      <version>${springboot.version}</version>
      <type>pom</type>
      <scope>import</scope>
    </dependency>
  </dependencies>
</dependencyManagement>
```
(and then declare all spring direct dependencies without a version under `dependencies`)

Unless ofc there's a particular need to override a dependency (but even then the version should be specified in the `dependencyManagement` section and the dependency declared without a version -- so it inherits the one define in the `dependencyManagement`)

yoti-sdk-parent/pom.xml
```diff
@@ -143,7 +145,7 @@
       <dependency>
         <groupId>com.fasterxml.jackson.core</groupId>
         <artifactId>jackson-databind</artifactId>
-        <version>${jackson-databind.version}</version>
+        <version>${jackson.version}</version>
       </dependency>
```
pn-santos, line 149: Theres a bom for jackson:
```xml
<dependency>
  <groupId>com.fasterxml.jackson</groupId>
  <artifactId>jackson-bom</artifactId>
  <version>${jackson.version}</version>
  <scope>import</scope>
  <type>pom</type>
</dependency>
```

yoti-sdk-parent/pom.xml
irotech: thanks for point it out. Since v3.0.0 the minimum required is Java 8 so we're good with this change

pn-santos (approved):

MrBurtyyy (approved):

bentaye (approved):

PR #362, opened 2023-02-20
SDK-2231: Add Identity Session creation service

irotech:
The PR seems big but if excluded the duplicated classes, Extensions, Policy, in common with the Dynamic Scenario the new code isn't that much. I had to duplicate those classes because I cannot move them from their original package and neither I want to use a DynamicPolicy obj in the v2 share. When v1 is going to be deprecated the whole package can be dropped. 

The little example just displays the result session id, status and expiration in page.
Next step Qr code creation and the example will show the resulting session and QR code

### DIFF
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/identity/ShareSessionNotification.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/identity/ShareSessionNotification.java
@@ -1,0 +1,104 @@
+package com.yoti.api.client.identity;
+
+import static com.yoti.api.client.spi.remote.util.Validation.notNullOrEmpty;
+
+import java.net.URI;
+import java.util.HashMap;
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public final class ShareSessionNotification {
+
+    @JsonProperty(Property.URL)
+    private final String url;
+
+    @JsonProperty(Property.METHOD)
+    private final String method;
+
+    @JsonProperty(Property.VERIFY_TLS)
+    private final boolean verifyTls;
+
+    @JsonProperty(Property.HEADERS)
+    private final Map<String, String> headers;
+
+    private ShareSessionNotification(Builder builder) {
+        url = builder.url;
+        method = builder.method;
+        verifyTls = builder.verifyTls;
+        headers = builder.headers;
+    }
+
+    public String getUrl() {
+        return url;
+    }
+
+    public String getMethod() {
+        return method;
+    }
+
+    public boolean isVerifyTls() {
+        return verifyTls;
+    }
+
+    public Map<String, String> getHeaders() {
+        return headers;
+    }
+
+    public static Builder builder(URI uri) {
+        return new Builder(uri);
+    }
+
+    public static final class Builder {
+
+        private final String url;
+        private final Map<String, String> headers;
+
+        private String method;
+        private boolean verifyTls;
+
+        private Builder(URI uri) {
+            url = uri.toString();
+            method = "POST";
+            verifyTls = true;
+            headers = new HashMap<>();
+        }
+
+        public Builder withMethod(String method) {
+            this.method = method;
+            return this;
+        }
+
+        public Builder withVerifyTls(boolean verifyTls) {
+            this.verifyTls = verifyTls;
+            return this;
+        }
+
+        public Builder withHeaders(Map<String, String> headers) {
+            this.headers.putAll(headers);
+            return this;
+        }
+
+        public Builder withHeader(String key, String value) {
+            headers.put(key, value);
+            return this;
+        }
+
+        public ShareSessionNotification build() {
+            notNullOrEmpty(url, Property.URL);
+
+            return new ShareSessionNotification(this);
+        }
+
+    }
+
+    private static final class Property {
+
+        private static final String URL = "url";
+        private static final String METHOD = "method";
+        private static final String VERIFY_TLS = "verifyTls";
+        private static final String HEADERS = "headers";
+
+    }
+
+}
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/identity/constraint/PreferredSources.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/identity/constraint/PreferredSources.java
@@ -1,0 +1,37 @@
+package com.yoti.api.client.identity.constraint;
+
+import java.util.List;
+
+import com.yoti.api.client.identity.policy.WantedAnchor;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class PreferredSources {
+
+    @JsonProperty(Property.ANCHORS)
+    private final List<WantedAnchor> wantedAnchors;
+
+    @JsonProperty(Property.SOFT_PREFERENCE)
+    private final boolean softPreference;
+
+    PreferredSources(List<WantedAnchor> wantedAnchors, boolean softPreference) {
+        this.wantedAnchors = wantedAnchors;
+        this.softPreference = softPreference;
+    }
+
+    public List<WantedAnchor> getWantedAnchors() {
+        return wantedAnchors;
+    }
+
+    public boolean isSoftPreference() {
+        return softPreference;
+    }
+
+    private static final class Property {
+
+        private static final String ANCHORS = "anchors";
+        private static final String SOFT_PREFERENCE = "soft_preference";
+
+    }
+
+}
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/identity/constraint/SourceConstraint.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/identity/constraint/SourceConstraint.java
@@ -1,0 +1,74 @@
+package com.yoti.api.client.identity.constraint;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import com.yoti.api.client.identity.policy.WantedAnchor;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class SourceConstraint implements Constraint {
+
+    @JsonProperty(Property.TYPE)
+    private final String type;
+
+    @JsonProperty(Property.PREFERRED_SOURCES)
+    private final PreferredSources preferredSources;
+
+    SourceConstraint(List<WantedAnchor> wantedAnchors, boolean softPreference) {
+        type = "SOURCE";
+        preferredSources = new PreferredSources(wantedAnchors, softPreference);
+    }
+
+    @Override
+    public String getType() {
+        return type;
+    }
+
+    public PreferredSources getPreferredSources() {
+        return preferredSources;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
+
+        private List<WantedAnchor> wantedAnchors;
+        private boolean softPreference;
+
+        private Builder() {
+            this.wantedAnchors = new ArrayList<>();
+        }
+
+        public Builder withWantedAnchors(List<WantedAnchor> wantedAnchors) {
+            this.wantedAnchors = Collections.unmodifiableList(wantedAnchors);
+            return this;
+        }
+
+        public Builder withWantedAnchor(WantedAnchor wantedAnchor) {
+            this.wantedAnchors.add(wantedAnchor);
+            return this;
+        }
+
+        public Builder withSoftPreference(boolean softPreference) {
+            this.softPreference = softPreference;
+            return this;
+        }
+
+        public SourceConstraint build() {
+            return new SourceConstraint(Collections.unmodifiableList(wantedAnchors),  softPreference);
+        }
+
+    }
+
+    private static final class Property {
+
+        private static final String TYPE = "type";
+        private static final String PREFERRED_SOURCES = "preferred_sources";
+
+    }
+
+}
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/identity/policy/WantedAttribute.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/identity/policy/WantedAttribute.java
@@ -1,0 +1,122 @@
+package com.yoti.api.client.identity.policy;
+
+import static com.yoti.api.client.spi.remote.util.Validation.notNullOrEmpty;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import com.yoti.api.client.identity.constraint.Constraint;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class WantedAttribute {
+
+    @JsonProperty(Property.NAME)
+    private final String name;
+
+    @JsonProperty(Property.DERIVATION)
+    private final String derivation;
+
+    @JsonProperty(Property.OPTIONAL)
+    private final boolean optional;
+
+    @JsonProperty(Property.ACCEPT_SELF_ASSERTED)
+    private final Boolean acceptSelfAsserted;
+
+    @JsonProperty(Property.CONSTRAINTS)
+    private final List<Constraint> constraints;
+
+    private WantedAttribute(Builder builder) {
+        name = builder.name;
+        derivation = builder.derivation;
+        optional = builder.optional;
+        acceptSelfAsserted = builder.acceptSelfAsserted;
+        constraints = builder.constraints;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public String getDerivation() {
+        return derivation;
+    }
+
+    public boolean isOptional() {
+        return optional;
+    }
+
+    public Boolean getAcceptSelfAsserted() {
+        return acceptSelfAsserted;
+    }
+
+    public List<Constraint> getConstraints() {
+        return constraints;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
+
+        private String name;
+        private String derivation;
+        private boolean optional;
+        private Boolean acceptSelfAsserted;
+        private List<Constraint> constraints;
+
+        private Builder() {
+            this.constraints = new ArrayList<>();
+        }
+        
+        public Builder withName(String name) {
+            this.name = name;
+            return this;
+        }
+        
+        public Builder withDerivation(String derivation) {
+            this.derivation = derivation;
+            return this;
+        }
+        
+        public Builder withOptional(boolean optional) {
+            this.optional = optional;
+            return this;
+        }
+
+        public Builder withAcceptSelfAsserted(boolean acceptSelfAsserted) {
+            this.acceptSelfAsserted = acceptSelfAsserted;
+            return this;
+        }
+
+        public Builder withConstraints(List<Constraint> constraints) {
+            this.constraints = Collections.unmodifiableList(constraints);
+            return this;
+        }
+
+        public Builder withConstraint(Constraint constraint) {
+            this.constraints.add(constraint);
+            return this;
+        }
+
+        public WantedAttribute build() {
+            notNullOrEmpty(name, Property.NAME);
+
+            return new WantedAttribute(this);
+        }
+
+    }
+
+    private static final class Property {
+
+        private static final String NAME = "name";
+        private static final String DERIVATION = "derivation";
+        private static final String OPTIONAL = "optional";
+        private static final String ACCEPT_SELF_ASSERTED = "accept_self_asserted";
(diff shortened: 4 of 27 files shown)


yoti-sdk-api/src/main/java/com/yoti/api/client/identity/NotificationRepresentation.java
```diff
@@ -0,0 +1,109 @@
...
+        return method;
+    }
+
+    public boolean isVerifyTls() {
+        return verifyTls;
+    }
+
+    public Map<String, String> getHeaders() {
+        return headers;
+    }
```
pn-santos, line 47: Does this needs getters? Isn't this class used to serialisation only?

Same question applies to other representations. Representations which the client needs to _produce_ but not _consume_  should not have getters.

Representations that the client only _consumes_ should have their builders private (so only jackson can use them).

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/NotificationRepresentation.java
```diff
@@ -0,0 +1,109 @@
...
+
+    public Builder builder() {
+        return new Builder();
+    }
```
pn-santos, line 51: Maybe make it so this requires the URI? Out of all the properties it's the only thing that is actually "required" (everything else has defaults). It would be one less builder setter method (slightly less verbose).

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/NotificationRepresentation.java
```diff
@@ -0,0 +1,109 @@
...
+        private String method;
+        private boolean verifyTls;
+
+        private Builder() {
+            method = "POST";
+            verifyTls = true;
+            headers = new HashMap<>();
+        }
+
+        public Builder withUrl(String url) {
```
pn-santos, line 67: Argument should be `URI`

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/NotificationRepresentation.java
```diff
@@ -0,0 +1,109 @@
...
+    @JsonProperty(Property.METHOD)
+    private final String method;
+
+    @JsonProperty(Property.VERIFY_TLS)
+    private final boolean verifyTls;
+
+    @JsonProperty(Property.HEADERS)
+    private final Map<String, String> headers;
+
+    NotificationRepresentation(String url, String method, boolean verifyTls, Map<String, String> headers) {
```
pn-santos, line 24:
```java
NotificationRepresentation(Builder builder)
```

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/SessionCreation.java
```diff
@@ -0,0 +1,132 @@
...
+    private final String redirectUri;
+
+    @JsonProperty(Property.NOTIFICATION)
+    private final NotificationRepresentation notification;
+
+    SessionCreation(Map<String, Object> subject,
+            Policy policy,
+            List<Extension<?>> extensions,
+            String redirectUri,
+            NotificationRepresentation notification) {
```
pn-santos, line 36:
```java
SessionCreation(Builder builder)
```

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/policy/WantedAttribute.java
```diff
@@ -0,0 +1,145 @@
...
+
+    private static final class Property {
+
+        private static final String NAME = "name";
+        private static final String DERIVATION = "derivation";
+        private static final String OPTIONAL = "optional";
+        private static final String ACCEPT_SELF_ASSERTED = "accept_self_asserted";
+        private static final String CONSTRAINTS = "constraints";
+
+        private Property() { }
```
pn-santos, line 141: (nit) Since these `Property` classes are `private` ones, I think the ctor could be dropped - it's no like the outer class is going to be doing `new Property` (it's pretty obvious its a constants class).

yoti-sdk-api/src/main/java/com/yoti/api/client/YotiClient.java
```diff
@@ -116,6 +124,10 @@ public ShareUrlResult createShareUrl(DynamicScenario dynamicScenario) throws Dyn
         return dynamicSharingService.createShareUrl(appId, keyPair, dynamicScenario);
     }
 
+    public SessionResult createSession(SessionCreation sessionCreation) throws IdentityException {
```
pn-santos, line 127:
```
createSession(SessionCreation request)
```
?

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/constraint/SourceConstraint.java
```diff
@@ -10,14 +10,16 @@
 
 public class SourceConstraint implements Constraint {
 
+    private static final String SOURCE_CONSTRAINT = "SOURCE";
```
pn-santos, line 13:
```
TYPE = "SOURCE"
```
the class is already named `SourceConstraint` so the `SOURCE` value name is relative to `SourceConstraint`.

`SourceConstraint.SOURCE_CONSTRAINT` is obviously not ideal, `SourceConstraint.TYPE` is much nicer.

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/constraint/SourceConstraint.java
```diff
@@ -10,14 +10,16 @@
...
+
     @JsonProperty(Property.TYPE)
-    private static final String type = ConstraintConstants.SOURCE_CONSTRAINT;
+    private final String type = SOURCE_CONSTRAINT;
```
pn-santos, line 16: Avoid field initialisation mixed with ctor initialisation. Assign the value within the ctor (as it's easier to read if everything is initialised in the same "place")

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/identity/IdentityException.java
```diff
@@ -0,0 +1,21 @@
+package com.yoti.api.client.spi.remote.call.identity;
+
+public class IdentityException extends Exception {
```
pn-santos, line 3: What's the benefit of declaring this as a checked exception? We will be forcing the client to wrap the call in a try/catch, but the exception doesn't provide anything beyond what you'd get from a standard runtime exception.

As long as we document that we throw `IdentityException` the client is then free to catch it or let it bubble up. But if we make it checked we _force_ it to always catch it.

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/identity/IdentityService.java
```diff
@@ -0,0 +1,98 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class IdentityService {
+
+    public static IdentityService newInstance() {
+        return new IdentityService(
+                new ObjectMapper(),
```
pn-santos, line 33: The `ObjectMapper` should be configured so that it only considers explicitly declared `JsonProperty` annotated fields, to avoid "accidental" (de)serialisation due to a getter or setter for example.

It should also probably be configured to ignore `null` fields (go look at the config we use in connect's `JsonConfig`)

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/identity/IdentityService.java
```diff
@@ -0,0 +1,98 @@
...
+                new UnsignedPathFactory(),
+                new SignedRequestBuilderFactory()
+        );
+    }
+
+    static final Logger LOG = LoggerFactory.getLogger(IdentityService.class);
```
pn-santos, line 39: shouldn't this be `private`?

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/identity/IdentityService.java
```diff
@@ -0,0 +1,98 @@
...
+import com.yoti.api.client.spi.remote.call.SignedRequestBuilderFactory;
+import com.yoti.api.client.spi.remote.call.factory.UnsignedPathFactory;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class IdentityService {
+
+    public static IdentityService newInstance() {
```
pn-santos, line 31: public static methods should be declared _after_ the ctor - or does this project use a different setup?

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/identity/IdentityService.java
```diff
@@ -0,0 +1,98 @@
...
+    static final Logger LOG = LoggerFactory.getLogger(IdentityService.class);
+
+    private final ObjectMapper mapper;
+    private final UnsignedPathFactory pathFactory;
+    private final SignedRequestBuilderFactory requestBuilderFactory;
+
+    private final String apiUrl;
+
+    public IdentityService(
+            ObjectMapper mapper,
```
pn-santos, line 48: Not sure it's a good idea to let a client pass in an `ObjectMapper`... we should be in full control of how the (de)serialization is done. If we let a client control this then they could (by accident) "break" (de)serialization.

I don't see a reason why this should be exposed.

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/SessionResult.java
```diff
@@ -0,0 +1,38 @@
+package com.yoti.api.client.identity;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class SessionResult {
```
pn-santos, line 5: This should just be called `Session` (or `ShareSession`)

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/SessionCreation.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import com.yoti.api.client.identity.extension.Extension;
+import com.yoti.api.client.identity.policy.Policy;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class SessionCreation {
```
pn-santos, line 15: Maybe name it `ShareSessionRequest` ?

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/NotificationRepresentation.java
```diff
@@ -0,0 +1,109 @@
+package com.yoti.api.client.identity;
+
+import static com.yoti.api.client.spi.remote.util.Validation.notNullOrEmpty;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public final class NotificationRepresentation {
```
pn-santos, line 10: Maybe name this `NotificationConfig` or `PushNotification` or `ShareSessionPushNotification` or `ShareSessionNotification`


---

Because docs already has "Session" it might be a good idea to name the share session `ShareSession` instead of simply `Session`. And maybe use that prefix on all "share" related stuff.

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/constraint/ConstraintConstants.java
```diff
@@ -0,0 +1,7 @@
+package com.yoti.api.client.identity.constraint;
+
+public class ConstraintConstants {
+
+    public static final String SOURCE_CONSTRAINT = "SOURCE";
+
+}
```
pn-santos: I don't get this one... I though you moved the value within the `SourceConstraint` class... why have it declared here too?

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/constraint/PreferredSources.java
```diff
@@ -0,0 +1,39 @@
+package com.yoti.api.client.identity.constraint;
+
+import java.util.List;
+
+import com.yoti.api.client.identity.policy.WantedAnchor;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class PreferredSources {
```
pn-santos, line 9: I assume this will never be used for de-serialization? Cause as it stands, they don't support it. They only are good for serialisation.

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/identity/IdentityService.java
```diff
@@ -0,0 +1,98 @@
...
+        this.apiUrl = System.getProperty(PROPERTY_YOTI_API_URL, DEFAULT_IDENTITY_URL);
+    }
+
+    public SessionResult createSession(String appId, KeyPair keyPair, SessionCreation sessionCreation)
+            throws IdentityException {
+        notNull(appId, "Application id");
+        notNull(keyPair, "Application key Pair");
+        notNull(sessionCreation, "Session creation");
+
+        LOG.info("Requesting Session Creation");
```
pn-santos, line 64: This log message is not good... it has no dynamic info, which means out of 1000 requests _all_ of it's instances will be identical, and on top of that it's logged at `info` (and saying a request has been sent is at _most_ a `debug` if not `trace` level message).

Imagine being an integrator and ask yourself if you'd like to see this log message being logged over and over... how useful would that be? Try to log only messages that you could see yourself as an integrator being interested in seeing.

And avoid logging messages that cannot be distinguished between executions (i.e. messages with no runtime-specific information)

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/NotificationRepresentation.java
irotech: none of these representations had getters when I pushed. I had a ton of failures from PMD so I had to choose if suppress the error or add the getters. Having public getters seems the approach used everywhere even if I do not see the integrator creating an obj and use any of the getters before passing the obj as arg to the client.

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/factory/UnsignedPathFactory.java
```diff
@@ -19,6 +19,7 @@ public class UnsignedPathFactory {
     static final String DOCS_NEW_FACE_CAPTURE_RESOURCE_PATH_TEMPLATE = "/sessions/%s/resources/face-capture?sdkId=%s";
     static final String DOCS_UPLOAD_FACE_CAPTURE_IMAGE_PATH_TEMPLATE = "/sessions/%s/resources/face-capture/%s/image?sdkId=%s";
     static final String DOCS_TRIGGER_IBV_NOTIFICATION_PATH_TEMPLATE = "/sessions/%s/instructions/email?sdkId=%s";
+    static final String IDENTITY_SESSION_CREATION_TEMPLATE = "/v2/sessions";
```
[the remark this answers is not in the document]
irotech: did u mean `private`? It is already `final`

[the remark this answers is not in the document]
bentaye: sorry yes, I meant `private`

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/NotificationRepresentation.java
pn-santos: ok

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/identity/IdentityService.java
irotech: setup!? This is a jungle!

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/ShareSessionNotification.java
```diff
@@ -0,0 +1,104 @@
...
+    private final boolean verifyTls;
+
+    @JsonProperty(Property.HEADERS)
+    private final Map<String, String> headers;
+
+    private ShareSessionNotification(Builder builder) {
+        this.url = builder.url;
+        this.method = builder.method;
+        this.verifyTls = builder.verifyTls;
+        this.headers = builder.headers;
```
pn-santos, line 29: (nit) `this` can be dropped (applies elsewhere)

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/constraint/SourceConstraint.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.Collections;
+import java.util.List;
+
+import com.yoti.api.client.identity.policy.WantedAnchor;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class SourceConstraint implements Constraint {
+
+    private static final String TYPE = "SOURCE";
```
pn-santos, line 13: (nit) since this is only used once maybe inline it? Just a thought - no biggie if you wanna keep it as a constant.

yoti-sdk-api/src/main/java/com/yoti/json/ResourceMapper.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonSetter;
+import com.fasterxml.jackson.annotation.Nulls;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.cfg.MapperConfig;
+import com.fasterxml.jackson.databind.introspect.VisibilityChecker;
+
+public class ResourceMapper {
```
pn-santos, line 11: `final`

yoti-sdk-api/src/main/java/com/yoti/json/ResourceMapper.java
```diff
@@ -0,0 +1,36 @@
...
+    private static VisibilityChecker<?> configureVisibility(MapperConfig<?> config) {
+        return config.getDefaultVisibilityChecker()
+                .withFieldVisibility(JsonAutoDetect.Visibility.NONE)
+                .withGetterVisibility(JsonAutoDetect.Visibility.NONE)
+                .withIsGetterVisibility(JsonAutoDetect.Visibility.NONE)
+                .withSetterVisibility(JsonAutoDetect.Visibility.NONE)
+                .withCreatorVisibility(JsonAutoDetect.Visibility.NONE);
+    }
+
+    public static ObjectMapper mapper() {
```
pn-santos, line 32:
```java
ResourceMapper.mapper()
```
stutters... maybe `getInstance()` or simply move the `writeValueAsString` (and any other `ObjectMapper` methods used) here and make them static methods so one does `ResourceMapper.doSomething(,,,)`

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/identity/IdentityService.java
```diff
@@ -0,0 +1,96 @@
...
+        return new IdentityService(new UnsignedPathFactory(), new SignedRequestBuilderFactory());
+    }
+
+    public ShareSession createShareSession(String appId, KeyPair keyPair, ShareSessionRequest shareSessionRequest)
+            throws IdentityException {
+        notNullOrEmpty(appId, "Application id");
+        notNull(keyPair, "Application key Pair");
+        notNull(shareSessionRequest, "Share Session request");
+
+        LOG.info("Requesting Share Session Creation for SDK ID '{}", appId);
```
pn-santos, line 57: This still is going to be "spam" for a RP as it's very likely that they will always use the same App ID (so every log message is going to look the same)... It's better than no dynamic info but I feel like it should probably be a `debug` or even `trace` log level message.

pn-santos (approved):
Mostly nits but I think there's still a couple that should be addressed - but I don't want to be a blocker for this so if you wanna merge as is I'll pre-approve.

bentaye (approved):

PR #363, opened 2023-02-23
SDK-2249: Add Identity QR code creation

irotech:


yoti-sdk-api/src/main/java/com/yoti/api/client/YotiClient.java
```diff
@@ -141,6 +143,12 @@ public ShareSession createShareSession(ShareSessionRequest request) throws Ident
         return identityService.createShareSession(appId, keyPair, request);
     }
 
+    public ShareSessionQrCode createShareSessionQrCode(String sessionId, ShareSessionQrRequest request)
+            throws IdentityException {
+        LOG.debug("Requesting a Share Session QR code creation...");
```
pn-santos, line 148: Referring back to my previous comment. A log message without dynamic params that is bound to executed multiple times per minute/second is just going to become "spam" in logs.

At least include the `sessionId` in the message.

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/ShareSessionQrRequest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class ShareSessionQrRequest {
+
+    @JsonProperty(Property.TRANSPORT)
+    private final String transport;
+
+    @JsonProperty(Property.DISPLAY_MODE)
+    private final String displayMode;
```
pn-santos, line 11: I think @laurent-yoti has decided to leave this out of the SDKs for now. The service will use default values anyway (neither of these are mandatory) so probably best to keep SDKs aligned and drop thus for now (unless there's a compeling reason why they should be included)

cc\ @AdrienVioletYoti

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/identity/IdentityService.java
```diff
@@ -74,9 +76,37 @@ public ShareSession createShareSession(String appId, KeyPair keyPair, ShareSessi
...
-    byte[] serializePayload(ShareSessionRequest shareSessionRequest) throws JsonProcessingException {
+    public ShareSessionQrCode createShareQrCode(
+            String appId,
+            KeyPair keyPair,
+            String sessionId,
+            ShareSessionQrRequest shareSessionQrRequest) {
+        notNullOrEmpty(sessionId, "Session id");
+        notNull(shareSessionQrRequest, "Share Session QR code request");
+
+        LOG.info("Requesting Share Session QR code Creation for session ID '{}", sessionId);
```
pn-santos, line 87: There's a `Requesting a Share Session QR code creation...` in `YotiClient` (and logged at `debug`) that seems to be about exactly the same thing as this one, except this one is at `info` and actually has _more_ info.

You wanna avoid logging messages pertaining to the same "action/operation" that include the same info. It's fine to have them at different levels if they include different contextual information (with finer log messages including _more_ detail the the ones at coarser levels).

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/identity/IdentityService.java
```diff
@@ -74,9 +76,37 @@ public ShareSession createShareSession(String appId, KeyPair keyPair, ShareSessi
...
+
+        String path = pathFactory.createIdentitySessionQrCodePath(sessionId);
+
+        try {
+            byte[] payload = serializePayload(shareSessionQrRequest);
+            SignedRequest request = createSignedRequest(appId, keyPair, path, payload);
```
pn-santos, line 93: maybe inline `serializePayload`?

yoti-sdk-spring-boot-example/src/main/java/com/yoti/api/examples/springboot/YotiLoginController.java
```diff
@@ -212,9 +214,13 @@ public String identitySession(Model model) throws URISyntaxException {
...
+
+        ShareSession session = null;
+        ShareSessionQrCode sessionQrCode = null;
         try {
-            result = client.createShareSession(shareSessionRequest);
+            session = client.createShareSession(shareSessionRequest);
+            sessionQrCode = client.createShareSessionQrCode(session.getId(), shareSessionQrRequest);
         } catch (Exception ex) {
             LOG.error(ex.getMessage());
         }
```
pn-santos, line 226: When doing assignments that need to be wrapped with a try/catch it's cleaner (easier to read and potentially to re-use) if the try/catch is extracted to a private method so that the try/catch is split from the assignment:
```java
ShareSession session = createShareSession(shareSessionRequest);
ShareSessionQrCode sessionQrCode = createShareSessionQrCode(session, shareSessionQrRequest);
```
and then do:
```java
private ShareSession createShareSession(...) {
    try {
        return ...
    } catch (...) {
        // handle exception
    }
}
```

That way the "technicalities" get pushed to a private method and the reader can "focus" on the business/main logic.

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/ShareSessionQrRequest.java
irotech: it was mentioned in the sync-up as well. I'll drop it.

laurent-yoti: We have not really mentioned per say in the sync up, but as we showed the implementation we did, this is indeed ignored right now, and we let Connect use its default value.

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/factory/UnsignedPathFactory.java
```diff
@@ -20,6 +20,7 @@ public class UnsignedPathFactory {
     private static final String DOCS_UPLOAD_FACE_CAPTURE_IMAGE_PATH_TEMPLATE = "/sessions/%s/resources/face-capture/%s/image?sdkId=%s";
     private static final String DOCS_TRIGGER_IBV_NOTIFICATION_PATH_TEMPLATE = "/sessions/%s/instructions/email?sdkId=%s";
     private static final String IDENTITY_SESSION_CREATION_TEMPLATE = "/v2/sessions";
```
bentaye, line 22: I just notice here but  `"/v2/sessions"` is not a template
And then I am thinking, do we really need the `_PATH_TEMPLATE` and `_TEMPLATE` prefix in any of these values? 
I'd just remove then all
```
    private static final String PROFILE = "/profile/%s?appId=%s";
    private static final String AML_CHECK = "/aml-check?appId=%s";
    private static final String QR_CODE_APP = "/qrcodes/apps/%s";
    private static final String DOCS_CREATE_SESSION = "/sessions?sdkId=%s";
    ....
```

irotech: I know it is ugly but I found it this way and I'm just sticking with it. The best approach here would be to create an enum. This is one of the many things that need to be changed at the next major release (breaking changes)

bentaye: Fine, I just mentioned it because it is a small self-contained change, as these are all private variables. 
#[Boy scout rule](https://biratkirat.medium.com/step-8-the-boy-scout-rule-robert-c-martin-uncle-bob-9ac839778385) :)

yoti-sdk-spring-boot-example/src/main/java/com/yoti/api/examples/springboot/DigitalIdentityController.java
```diff
@@ -65,20 +62,34 @@ public String identityShare(Model model) throws URISyntaxException {
...
-        ShareSession result = null;
+        ShareSession session;
         try {
-            result = client.createShareSession(shareSessionRequest);
+            session = client.createShareSession(shareSessionRequest);
         } catch (Exception ex) {
-            LOG.error(ex.getMessage());
+            model.addAttribute("error", ex.getMessage());
+            return "error";
+        }
```
pn-santos, line 71: (nit) See https://github.com/getyoti/yoti-java-sdk/pull/380#discussion_r1129261372 same thing applies here

pn-santos (approved):

pn-santos (approved):

PR #364, opened 2023-02-23
SDK-2220: Add support for supplying an ApplicantProfile resource at s…

MrBurtyyy:
…ession creation

### DIFF
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/resources/ResourceCreationContainer.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/resources/ResourceCreationContainer.java
@@ -1,0 +1,51 @@
+package com.yoti.api.client.docs.session.create.resources;
+
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Container for creating resources at session creation.
+ */
+public class ResourceCreationContainer {
+
+    private static final String APPLICANT_PROFILE = "applicant_profile";
+
+    @JsonProperty(APPLICANT_PROFILE)
+    private final Map<String, Object> applicantProfile;
+
+    private ResourceCreationContainer(Map<String, Object> applicantProfile) {
+        this.applicantProfile = applicantProfile;
+    }
+
+    public static ResourceCreationContainer.Builder builder() {
+        return new ResourceCreationContainer.Builder();
+    }
+
+    /**
+     * The applicant profile resource to be created.
+     *
+     * @return the applicant profile fields
+     */
+    public Map<String, Object> getApplicantProfile() {
+        return applicantProfile;
+    }
+
+    public static final class Builder {
+
+        private Map<String, Object> applicantProfile;
+
+        private Builder() {}
+
+        public Builder withApplicantProfile(Map<String, Object> applicantProfile) {
+            this.applicantProfile = applicantProfile;
+            return this;
+        }
+
+        public ResourceCreationContainer build() {
+            return new ResourceCreationContainer(applicantProfile);
+        }
+
+    }
+
+}
--- a/yoti-sdk-api/src/test/java/com/yoti/api/client/docs/session/create/SessionSpecTest.java
+++ b/yoti-sdk-api/src/test/java/com/yoti/api/client/docs/session/create/SessionSpecTest.java
@@ -13,6 +13,7 @@ import java.util.Map;
 import com.yoti.api.client.docs.session.create.check.RequestedDocumentAuthenticityCheck;
 import com.yoti.api.client.docs.session.create.check.RequestedLivenessCheck;
 import com.yoti.api.client.docs.session.create.filters.RequiredDocument;
+import com.yoti.api.client.docs.session.create.resources.ResourceCreationContainer;
 import com.yoti.api.client.docs.session.create.task.RequestedIdDocTextExtractionTask;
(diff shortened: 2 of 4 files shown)


yoti-sdk-api/src/test/java/com/yoti/api/client/docs/session/create/SessionSpecTest.java
```diff
@@ -271,18 +273,27 @@ public void buildWithSubject() throws IOException {
...
+    @Test
+    public void shouldBuildWithResourceContainer() {
+        SessionSpec sessionSpec = SessionSpec.builder()
+                .withResources(resourceCreationContainerMock)
+                .build();
+
+        assertThat(sessionSpec.getResources(), is(resourceCreationContainerMock));
+    }
+
+    private static final class TestProperty {
```
bentaye, line 288: I don't understand the renaming
I feel like it would make more sense to split them in 2 classes
```
    private static final class IdentityProperty {

        private static final String TYPE = "type";
        private static final String SCHEME = "scheme";
        private static final String OBJECTIVE = "objective";
        private static final String TRUST_FRAMEWORK = "trust_framework";

        private IdentityProperty() { }

    }

    private static final class SubjectProperty {

        private static final String SUBJECT_ID = "subject_id";

        private SubjectProperty() { }

    }
```

Or something that make sense. `TestProperty` sounds like constants specific to the tests

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/resources/ResourceCreationContainer.java
```diff
@@ -0,0 +1,57 @@
...
+
+        private Builder() {}
+
+        public Builder withApplicantProfile(Map<String, Object> applicantProfile) {
+            this.applicantProfile = applicantProfile;
+            return this;
+        }
+
+        public ResourceCreationContainer build() {
+            return new ResourceCreationContainer(applicantProfile);
```
irotech, line 44: it's good practice to have the class contructor with the builder as arg when using Builders

```
private ResourceCreationContainer(Builder builder) {
        applicantProfile = builder.applicantProfile;
}

public ResourceCreationContainer build() {
            return new ResourceCreationContainer(this);
}
```

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/resources/ResourceCreationContainer.java
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+    }
+
+    private static final class Property {
+
+        private static final String APPLICANT_PROFILE = "applicant_profile";
+
+        private Property() {}
```
irotech, line 53: can be dropped

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/resources/ResourceCreationContainer.java
MrBurtyyy: This isn't how it's done in the rest of the SDK, so I'd like to keep it similar as possible.  If this is something you think would be good to do, probably something worth doing in the next major and updating everything.

yoti-sdk-api/src/test/java/com/yoti/api/client/docs/session/create/SessionSpecTest.java
MrBurtyyy: The rename came from an accidental refactor, where Intellij complained that the class didn't match a specific regex.  Agree that we should probably split them out, I'll do that now :+1:

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/resources/ResourceCreationContainer.java
irotech: I'm afraid the next major release requires a nuke :)

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/resources/ResourceCreationContainer.java
```diff
@@ -0,0 +1,51 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Container for creating resources at session creation.
+ */
+public class ResourceCreationContainer {
+
+    private static final String APPLICANT_PROFILE = "applicant_profile";
+
+    @JsonProperty(APPLICANT_PROFILE)
```
irotech, line 14: in `SessionSpec` a Property class in here a constant, why different approach?

yoti-sdk-api/src/test/java/com/yoti/api/client/docs/session/create/SessionSpecTest.java
```diff
@@ -271,18 +273,34 @@ public void buildWithSubject() throws IOException {
...
+    }
+
+    private static final class SubjectProperty {
+
         private static final String SUBJECT_ID = "subject_id";
 
-        private Property() { }
+        private SubjectProperty() {}
 
     }
```
irotech, line 305: I'd personally have done 

```
class Property { 

    lass Identity { .. }
    class Subject { .. }
    
}
```
so u do `Property.Identity.fieldName`

irotech (approved):

PR #369, opened 2023-02-28
SDK-2154: Add support for requesting Static Liveness Check

artem-yoti:


yoti-sdk-api/src/test/java/com/yoti/api/client/docs/session/create/check/RequestedLivenessCheckTest.java
```diff
@@ -38,6 +38,18 @@ public void shouldBuildRequestedLivenessCheckWithZoomLivenessType() {
         assertThat(configResult.getMaxRetries(), is(MAX_RETRIES_VALUE));
     }
 
+    @Test
+    public void shouldBuildRequestedLivenessCheckWithStaticcLivenessType() {
```
MrBurtyyy, line 42:
```suggestion
    public void shouldBuildRequestedLivenessCheckWithStaticLivenessType() {
```

MrBurtyyy (commented):
Just the fix to the name of the test method, otherwise LGTM

MrBurtyyy (changes requested):
Actually, now that I think about it, you will also need to add support when fetching the session config.

Take a look at `RequiredLivenessResourceResponse`, and implement the required subclass for static liveness.

MrBurtyyy (approved):

PR #371, opened 2023-02-28
SDK-1889: Allow attempts configuration to be set at session creation

MrBurtyyy:


yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/AttemptsConfiguration.java
```diff
@@ -0,0 +1,75 @@
+package com.yoti.api.client.docs.session.create;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public final class AttemptsConfiguration {
+
+    @JsonProperty("ID_DOCUMENT_TEXT_DATA_EXTRACTION")
```
irotech, line 10: Property class?

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/AttemptsConfiguration.java
```diff
@@ -0,0 +1,75 @@
...
+    private final Map<String, Integer> idDocumentTextDataExtractionAttemptsConfig;
+
+    private AttemptsConfiguration(Map<String, Integer> idDocumentTextDataExtractionAttemptsConfig) {
```
irotech, line 13: reduce maybe verbosity with config as arg and drop `this`?

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/AttemptsConfiguration.java
```diff
@@ -0,0 +1,75 @@
...
+        private static final String GENERIC = "GENERIC";
+
+        private Map<String, Integer> idDocumentTextDataExtractionAttempts;
+
+        private Builder() {}
+
+        private Builder withIdDocumentTextExtractionCategoryRetries(String category, int numberOfRetries) {
+            if (idDocumentTextDataExtractionAttempts == null) {
+                idDocumentTextDataExtractionAttempts = new HashMap<>();
+            }
```
irotech, line 42: maybe move this to the constructor? Always initialize collections in the Builder constructor

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/AttemptsConfiguration.java
```diff
@@ -0,0 +1,75 @@
+package com.yoti.api.client.docs.session.create;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public final class AttemptsConfiguration {
```
irotech, line 8: don't know if u're constrained here but there are very long names in here

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/SdkConfig.java
```diff
@@ -39,6 +39,9 @@ public class SdkConfig {
     @JsonProperty("allow_handoff")
     private final Boolean allowHandoff;
 
+    @JsonProperty("attempts_configuration")
```
irotech, line 42: could u create a Property class in here as well?

irotech (approved):
only minor things

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/AttemptsConfiguration.java
MrBurtyyy: We don't want this though - that means we would always send an empty object when serialising to JSON.  If we add other property maps in here, we don't want to default the others to empty objects (rather we want to omit them).

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/AttemptsConfiguration.java
MrBurtyyy: I've reduced them down a little bit, but so that they still make sense to the integrator.

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/SdkConfig.java
MrBurtyyy: Done

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/AttemptsConfiguration.java
irotech: So u're saying that server side u want either no json property (null map) or not empty json property (map with values).

MrBurtyyy: Yes - it just ensures that the SDK is never opinionated, if the integrator doesn't call those methods for that specific property.

PR #375, opened 2023-03-01
SDK-2256: Add Digital Identity Session QR code retrieval service

irotech:


yoti-sdk-api/src/main/java/com/yoti/api/client/DigitalIdentityClient.java
```diff
@@ -73,11 +73,21 @@ public ShareSessionQrCode createShareQrCode(String sessionId) throws DigitalIden
...
-    public Object fetchShareQrCode(String qrCodeId) {
+    /**
+     * Retrieve the sharing session QR code
+     *
+     * @param qrCodeId
+     *              ID of the QR code to retrieve
+     * @return an {@link ShareSessionQrCode}
+     *              The content of the Share Session QR code
+     * @throws DigitalIdentityException
+     *              Aggregate exception signalling issues during the call
```
pn-santos, line 84: This is not a conventional javadoc format. The descriptions should be in the same line as the var names.

Also there no need to `{@link ShareSessionQrCode}` since the method signature already specifies the return type

And 
```
Aggregate exception signalling issues during the call
```
probably would be better as:
```
Thrown if the retrieval is unsuccessful
```

yoti-sdk-api/src/main/java/com/yoti/api/client/DigitalIdentityClient.java
```diff
@@ -73,11 +73,21 @@ public ShareSessionQrCode createShareQrCode(String sessionId) throws DigitalIden
...
+     */
+    public ShareSessionQrCode fetchShareQrCode(String qrCodeId) throws DigitalIdentityException {
         notNullOrEmpty(sdkId, "SDK ID");
         notNull(keyPair, "Application Key Pair");
```
pn-santos, line 89: Why do these checks here? Shouldn't these be done by the class ctor?

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/ShareSessionQrCode.java
```diff
@@ -10,6 +14,21 @@ public class ShareSessionQrCode {
...
+    private String policy;
+
+    @JsonProperty(Property.EXTENSIONS)
+    private List<Extension> extensions;
+
+    @JsonProperty(Property.SESSION)
+    private ShareSession session;
+
+    @JsonProperty(Property.REDIRECT_URI)
+    private String redirectUri;
```
pn-santos, line 30: It's always preferable to use strong types (in this case `URI`

pn-santos (approved):

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/ShareSessionQrCode.java
bentaye: maybe change this one too in the same class if possible
```
    @JsonProperty(Property.URI)
    private String uri;
```

yoti-sdk-spring-boot-example/src/main/java/com/yoti/api/examples/springboot/DigitalIdentityController.java
```diff
@@ -83,6 +84,11 @@ public String identityShareComplete(Model model) throws URISyntaxException {
         LOG.info("Requesting Share QR code creation for session with ID '{}'", sessionId);
         ShareSessionQrCode sessionQrCode = createShareQrCode(sessionId);
 
+        String qrCodeId = sessionQrCode.getId();
+
+        LOG.info("Requesting Share QR code with ID '{}'", qrCodeId);
```
bentaye, line 89: This log is repeated in `DigitalIdentityService.fetchShareQrCode` 
```
LOG.info("Requesting Share Session QR code with ID '{} at '{}'", qrCodeId, path);
```
so it is probably not needed here. 

does it have to be INFO, it seems like a DEBUG level log

Same thing for the log above: (I know this one wasn't added by you)
```
LOG.info("Requesting Share QR code creation for session with ID '{}'", sessionId);
```

pn-santos (approved):

PR #380, opened 2023-03-07
SDK-2262: Add Digital Identity Session Receipt retrieval service

irotech:


yoti-sdk-api/src/main/java/com/yoti/api/client/DigitalIdentityClient.java
```diff
@@ -81,8 +81,18 @@ public ShareSessionQrCode fetchShareQrCode(String qrCodeId) throws DigitalIdenti
         return identityService.fetchShareQrCode(sdkId, keyPair, qrCodeId);
     }
 
-    public Object fetchShareReceipt(String receiptId) {
-        return identityService.fetchShareReceipt(receiptId);
+    /**
+     * Retrieve the user profile, containing the user's defined attributes, from the share receipt
```
pn-santos, line 85: I think "user profile" no longer applies. For share v2 there's no "profile".

I think this would be more accurately phrased like so:
```
Retrieve the decrypted share receipt.

<p>A receipt will contain the shared user attributes.</p>
```

yoti-sdk-api/src/main/java/com/yoti/api/client/ReceiptDetails.java
```diff
@@ -0,0 +1,121 @@
...
+
+public final class ReceiptDetails {
+
+    private final String id;
+    private final String sessionId;
+    private final String rememberMeId;
+    private final String parentRememberMeId;
+    private final ApplicationProfile profile;
+    private final HumanProfile otherPartyProfile;
+    private final Date date;
```
pn-santos, line 14: Why use `Date`? `OffsetDateTime` should be used instead. The only reason to use `Date` would be if we wanted Java 7 compatibility

yoti-sdk-api/src/main/java/com/yoti/api/client/ReceiptDetails.java
```diff
@@ -0,0 +1,121 @@
...
+            return this;
+        }
+
+        public Builder otherPartyProfile(HumanProfile otherPartyProfile) {
+            this.otherPartyProfile = otherPartyProfile;
+            return this;
+        }
+
+        public Builder date(Date date) {
+            this.date = Optional.ofNullable(date).map(d -> new Date(d.getTime())).orElse(null);
```
pn-santos, line 106: Humm a receipt timestamp will never be `null`

yoti-sdk-api/src/main/java/com/yoti/api/client/ReceiptDetails.java
```diff
@@ -0,0 +1,121 @@
...
+        return parentRememberMeId;
+    }
+
+    public ApplicationProfile getProfile() {
+        return profile;
+    }
+
+    public HumanProfile getOtherPartyProfile() {
+        return otherPartyProfile;
+    }
```
pn-santos, line 50: All of these values are "optional", i.e. a receipt _might_ not contain them.

This would be better expressed if the return values were wrapped in `Optional`

Although from a technical perspective, given the way the `ReceiptDetails` instance is built the "profiles" won't be `null` they can be "empty" (as in they contain no data).

One option would be to only set them if they contain data (so if "empty" the value would be left `null`.

yoti-sdk-api/src/main/java/com/yoti/api/client/ReceiptDetails.java
```diff
@@ -0,0 +1,121 @@
...
+
+    public Date getDate() {
+        return Optional.ofNullable(date).map(d -> new Date(d.getTime())).orElse(null);
+    }
+
+    public ExtraData getExtraData() {
+        return extraData;
+    }
```
pn-santos, line 58: Same here regarding being "optional"

yoti-sdk-api/src/main/java/com/yoti/api/client/ReceiptDetails.java
```diff
@@ -0,0 +1,121 @@
+package com.yoti.api.client;
+
+import java.util.Date;
+import java.util.Optional;
+
+public final class ReceiptDetails {
```
pn-santos, line 6: This should be called only `Receipt`

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/ReceiptDetailsFactory.java
```diff
@@ -0,0 +1,107 @@
...
+
+        List<Attribute<?>> profileAttr = attributeListReader.read(wrappedReceipt.getProfile(), receiptKey);
+        List<Attribute<?>> otherPartyProfileAttr = attributeListReader.read(wrappedReceipt.getOtherPartyProfile(), receiptKey);
+        return ReceiptDetails.forReceipt(wrappedReceipt.getId())
+                .sessionId(wrappedReceipt.getSessionId())
+                .profile(new ApplicationProfile(profileAttr))
+                .otherPartyProfile(new HumanProfile(otherPartyProfileAttr))
+                .extraData(parseExtraData(wrappedReceipt.getExtraData(), receiptKey))
+                .rememberMeId(parseRememberMeId(wrappedReceipt.getRememberMeId()))
+                .parentRememberMeId(parseRememberMeId(wrappedReceipt.getParentRememberMeId()))
```
pn-santos, line 69: As mentioned, all of these _might_ not exist in a successful receipt. I think it would be preferable if this didn't set the values if there's no value to set (cause it wasn't in the receipt representation fetched).

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ShareSessionReceipt.java
```diff
@@ -0,0 +1,206 @@
+package com.yoti.api.client.spi.remote.call;
```
pn-santos, line 1: maybe `client` instead of `call`? Or even `api`?

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ShareSessionReceipt.java
```diff
@@ -0,0 +1,206 @@
...
+        return content.extraData();
+    }
+
+    public byte[] getOtherPartyProfile() {
+        return otherPartyContent.profile();
+    }
+
+    public byte[] getOtherPartyExtraData() {
+        return otherPartyContent.extraData();
+    }
```
pn-santos, line 60: As mentioned above, all of these _might_ not exist in a receipt. Make sure that this doesn't blow up if any of these are `null` in the response.

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ShareSessionReceipt.java
```diff
@@ -0,0 +1,206 @@
...
+        private static final String OTHER_PARTY_CONTENT = "otherPartyContent";
+        private static final String WRAPPED_ITEM_KEY_ID = "wrappedItemKeyId";
+        private static final String WRAPPED_KEY = "wrappedKey";
+
+    }
+
+    private static class Content {
+
+        private byte[] profile;
+        private byte[] extraData;
```
pn-santos, line 173: Same applies to these ones. they are "optional" and might not be present in the response.

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ShareSessionReceiptItemKey.java
```diff
@@ -0,0 +1,78 @@
...
+        return id;
+    }
+
+    public byte[] getIv() {
+        return Optional.ofNullable(iv).map(byte[]::clone).orElse(null);
+    }
+
+    public byte[] getValue() {
+        return Optional.ofNullable(value).map(byte[]::clone).orElse(null);
+    }
```
pn-santos, line 32: These will never be `null`

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ShareSessionReceiptItemKey.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Optional;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+@JsonDeserialize(builder = ShareSessionReceiptItemKey.Builder.class)
+public final class ShareSessionReceiptItemKey {
+
+    private final String id;
+    private final byte[] iv;
```
pn-santos, line 13: (nit) there's a better type for this value: `IvParameterSpec`

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/identity/DigitalIdentityService.java
```diff
@@ -146,8 +160,43 @@ public ShareSessionQrCode fetchShareQrCode(String sdkId, KeyPair keyPair, String
...
+        ShareSessionReceipt receipt;
+        try {
+            SignedRequest request = createSignedRequest(sdkId, keyPair, receiptPath);
+            receipt = request.execute(ShareSessionReceipt.class);
+        } catch (Exception ex) {
+            throw new DigitalIdentityException(
+                    String.format("Error while fetching the share session QR code with ID '{%s}' ", receiptId),
+                    ex
+            );
+        }
```
pn-santos, line 176: When doing assignments "wrapped" in a try/catch if makes for more readable code if the try/catch is extracted to a private method (so the try/catch always returns a value instead of assigning one), e.g.
```java
ShareSessionReceipt receipt = doFetchShareReceipt(sdkId, keyPair, receiptPath);
ShareSessionReceiptItemKey receiptKey = fetchShareReceiptKey(sdkId, keyPair, receipt);
```

Extracting the try/catch to a private method, removes the language "cruft" from the "main" logic making it easier to focus on the business logic (**what** is actually being done) instead of **how** it's being done).

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ShareSessionReceipt.java
irotech: following the current package structure, few things are shared and need to be cloned if I want to better structure the code. I opted to stick to the current (not very nice) structure rather than copying and pasting xxx classes

yoti-sdk-api/src/main/java/com/yoti/api/client/Receipt.java
```diff
@@ -56,6 +60,14 @@ public ExtraData getExtraData() {
         return extraData;
     }
 
+    public ExtraData getOtherPartyExtraData() {
+        return otherPartyExtraData;
+    }
+
+    public String getError() {
+        return error;
+    }
```
pn-santos, line 69: Regarding https://github.com/getyoti/yoti-java-sdk/pull/380#discussion_r1129238821 does returning an `Optional` wrapped value not be preferable for all the values that _might_ not exist?

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/ReceiptDetailsFactory.java
```diff
@@ -0,0 +1,104 @@
...
+import com.yoti.api.client.spi.remote.call.ShareSessionReceiptItemKey;
+import com.yoti.api.client.spi.remote.call.identity.DigitalIdentityException;
+import com.yoti.crypto.CipherType;
+import com.yoti.crypto.Crypto;
+import com.yoti.validation.Validation;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ReceiptDetailsFactory {
```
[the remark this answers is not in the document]
irotech: in the last commit I renamed it `ReceiptBuilderFactory` to align the change made from `ReceiptDetails` -> `Receipt`

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/ShareSessionReceipt.java
```diff
@@ -0,0 +1,208 @@
...
+import java.time.OffsetDateTime;
+import java.util.Base64;
+import java.util.Objects;
+import java.util.Optional;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+@JsonDeserialize(builder = ShareSessionReceipt.Builder.class)
+public final class ShareSessionReceipt {
```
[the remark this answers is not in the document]
irotech: I called it `ShareSessionReceipt` to differenciate it from the `Receipt` v1. I guess I can call both the same since there is no cross reference between them tho.

pn-santos (approved):

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/identity/Receipt.java
```diff
@@ -44,28 +45,28 @@ public String getParentRememberMeId() {
...
-        return extraData;
+    public ApplicationContent getApplicationContent() {
+        return applicationContent;
     }
 
-    public ExtraData getOtherPartyExtraData() {
-        return otherPartyExtraData;
+    public UserContent getUserContent() {
+        return userContent;
     }
```
irotech: @laurent-yoti @dhivya-pattabhiraman I made the change we discussed on Fri. This is the solution, among the suggested, I like the most. @jlmartyres is also looking at it.

laurent-yoti: Hum, I am a bit confused...

`getProfile()` : it is acknowledged that the RP using the SDK is interested mostly in that bit, which is the user profile, what you suggesting here is in line with the first proposal, I am only confused by the optional aspect of it... although, in theory, you are right, it could be empty if the user has only shared their `rememberMeId`.

`getExtraData()`: I believe that the outcome of our discussion and from what you saw in the code, these extra data is actually corresponding to the application content's extra data... so it should read `returns optional applicationExtraData`

`getApplicationContent()` --> happy to have that

`getUserContent()` --> happy to have that


@dhivya-pattabhiraman @jlmartyres @irotech ?

irotech: About the profile (`userProfile`), the SDK needs to account for the missing (null) values and the only field we can guarantee is never null is the `applicationProfile`. As you said a share with a `rememberMeId` only will return a null `userProfile`.

The `getExtraData()` bit is something I didn't quite get. Currently the SDK, in the dynamic scenario, returns the applicationExtradata only completely ignoring the userExtradata. So I asked myself shouldn't it be the extraData Core sets, in share completion, along with the userProfile? Means both the values in the user content?

laurent-yoti: Different taste/strategy for the profile, you could always have a profile with 0 attributes inside, so it would always exist (just a suggestion).

From what we saw in the current implementation of the receipt, there is only 1 `extraData` and it seemed to be linked with the application content, in the sense that it reads `extra_data_content` ... and we kinda concluded  that:
- `profile_content` => application/RP
- `extra_data_content`  => application/RP
- `other_party_profile_content`  => user
- `other_party_extra_data_content`  => user

We can give it a try in our test environment.

irotech: @pgreci can you please shed a light on these extraData please? We said both, application and user, have a meaning but there is no real use cases (yet). 
The Java SDK, dynamic scenario, returns both profiles but only the applicationExtraData which in Core language is translated in 
```
    optional string other_party_profile_uri = 2;
    optional string profile_uri = 3;
    optional string extra_data_uri = 5;
```

In v2 we want to create an API for the integrator to have a short 
`getProfile()` --> **otherPartyProfile** value
`getExtraData()` --> ??? shouldn't this be **otherPartyExtraData**? Or **extraData** as it is now?

and additional
`getApplicationContent()` (contains both **profile** and **extraData**)
`getUserContent()` (contains both **otherPartyProfile** and **otherPartyExtraData**)

laurent-yoti: @irotech I thought we saw that Connect mapping for the current v1 is returning only the `extra_data` and that this is mapped to the application content... as in, these are the additional data provided by the application (RP) to dictate some of the share operations (location for example).
I think we can quickly check that out on stg1, using both v1 and v2, and see what the extra_data of v1 matches with, user's or application's ones.

irotech: I did a test with this request

```
        Extension<LocationConstraintContent> locationExtension = new LocationConstraintExtensionBuilder()
                .withLatitude(51.5074)
                .withLongitude(-0.1278)
                .withRadius(6000)
                .build();

        Policy policy = Policy.builder()
                .withWantedRememberMe(true)
                .build();

        ShareSessionRequest shareSessionRequest = ShareSessionRequest.builder()
                .withPolicy(policy)
                .withRedirectUri(new URI("https://host/redirect/"))
                .withExtension(locationExtension)
                .build();
```

for reference the receipt ID; `cy7f1+00qvd1pKSBCzcbzVDCEl2w/jg4evu0Sa9fCbTkMKwVkclE1XijZlNWDVnv`

the wrapped receipt 
```
 id = "cy7f1+00qvd1pKSBCzcbzVDCEl2w/jg4evu0Sa9fCbTkMKwVkclE1XijZlNWDVnv"
 sessionId = "ss.v2.ChY2OHNvQ2RUU1RwT19kWU85dDVlZmhBEgdsZDUuZ2Jy"
 timestamp = {OffsetDateTime@9160} "2023-03-13T16:07:07Z"
 content = {WrappedReceipt$Content@9161} 
  profile = {byte[245]@9168} [10, ... -108, +145 more]
  extraData = null
 otherPartyContent = {WrappedReceipt$Content@9162} 
  profile = null
  extraData = {byte[68]@9169} [10, ... 42, -90] <-- otherPartyContent.extraData has the location value
 wrappedItemKeyId = "218bd625-cb7e-4f3a-a5aa-12c1fde2101f"
 rememberMeId = {byte[48]@9164} [97, ... 113, -77]
 parentRememberMeId = {byte[48]@9165} [-32, ... -92, -89]
 wrappedKey = {byte[48]@9166} [-54, ... 51, -66]
 error = null
```

there is no converter if not for THIRD_PARTY_ATTRIBUTE in the Java SDK so the extraData value for LOCATION is lost

The final receipt came out, as I expected, this way:

```
 id = "cy7f1+00qvd1pKSBCzcbzVDCEl2w/jg4evu0Sa9fCbTkMKwVkclE1XijZlNWDVnv"
 sessionId = "ss.v2.ChY2OHNvQ2RUU1RwT19kWU85dDVlZmhBEgdsZDUuZ2Jy"
 rememberMeId = "YZVN6sW3pbh/f91KXerUk2OrLOhZMFk/nQCBzTwFIa80sb5tAJtTZn/KC72+nXGz"
 parentRememberMeId = "4PpmWo0toxWo5qvnz/kDu7g81LNdo5j2LttLQpffFEnUr/u+cmjfNA2ztznm/qSn"
 timestamp = {OffsetDateTime@9205} "2023-03-13T16:07:07Z"
 applicationContent = {Receipt$ApplicationContent@9230} 
  profile = {ApplicationProfile@9234} 
   protectedAttributes = {Collections$UnmodifiableMap@9235}  size = 4
    "application_logo" -> {ArrayList@9243}  size = 1
    "application_name" -> {ArrayList@9245}  size = 1
    "application_url" -> {ArrayList@9247}  size = 1
    "application_receipt_bgcolor" -> {ArrayList@9249}  size = 1
  extraData = null
 userContent = {Receipt$UserContent@9231} 
  profile = null
  extraData = {ExtraData@9233}  <-- the otherPartyExtraData (user)
   attributeIssuanceDetails = null
 error = null
```

yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/identity/DigitalIdentityService.java
```diff
@@ -155,18 +196,18 @@ SignedRequest createSignedRequest(String sdkId, KeyPair keyPair, String path)
...
 
-        Optional.ofNullable(payload).map(v ->
-                requestBuilder.withPayload(v).withHeader(CONTENT_TYPE, CONTENT_TYPE_JSON)
-        );
+        if (null != payload) {
+            request.withPayload(payload).withHeader(CONTENT_TYPE, CONTENT_TYPE_JSON);
+        }
 
-        return requestBuilder.build();
+        return request.build();
```
pn-santos, line 210: (only a suggestion)
```java
return Optional.ofNullable(payload)
        .map(request::withPayload)
        .map(r -> r.withHeader(..., ...))
        .orElse(request)
        .build()
```
If you rather keep the `if` I'd at least change it so it's more natural: 
```java
if (payload != null) {
    (...)
}
```

pn-santos (approved):

PR #386, opened 2023-03-30
Sdk 2196 - Add support for requesting the profile preview

artem-yoti:
Ticket - https://lampkicking.atlassian.net/browse/SDK-2196

### DIFF
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/SessionSpec.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/SessionSpec.java
@@ -59,6 +59,9 @@ public class SessionSpec {
     @JsonProperty(Property.RESOURCES)
     private final ResourceCreationContainer resources;
 
+    @JsonProperty(Property.CREATE_IDENTITY_PROFILE_PREVIEW)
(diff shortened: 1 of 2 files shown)


yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/SessionSpec.java
```diff
@@ -439,6 +465,7 @@ private static final class Property {
         private static final String IDENTITY_PROFILE_REQUIREMENTS = "identity_profile_requirements";
         private static final String SUBJECT = "subject";
         private static final String RESOURCES = "resources";
+        private static final String IDENTITY_PROFILE_PREVIEW = "create_identity_profile_preview";
```
MrBurtyyy, line 468: Can you set the property name to match the string please?

MrBurtyyy (changes requested):

yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/SessionSpec.java
artem-yoti: Done

MrBurtyyy (approved):

PR #413, opened 2023-07-25
SDK-2230: Expose share v2 API

irotech:
- Add support to Share V2


yoti-sdk-api/src/main/java/com/yoti/api/client/spi/remote/call/identity/ReceiptItemKey.java
```diff
@@ -0,0 +1,79 @@
...
+
+        @JsonProperty(Property.ID)
+        public Builder id(String id) {
+            this.id = id;
+            return this;
+        }
+
+        @JsonProperty(Property.IV)
+        public Builder iv(String iv) {
+            this.iv = new IvParameterSpec(decode(iv));
```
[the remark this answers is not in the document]
irotech: this is not an issue as it is used to decrypt

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/extension/ThirdPartyAttributeExtensionBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public Extension<ThirdPartyAttributeContent> build() {
+        ThirdPartyAttributeContent thirdPartyAttributeContent = new ThirdPartyAttributeContent(expiryDate, definitions);
+        return new Extension<>(TYPE, thirdPartyAttributeContent);
+    }
+
+    private static final class Property {
+
+        private static final String EXPIRY_DATE = "expiryDate";
```
pn-santos, line 49: Just to double check, is the use of camelCase here intentional? in `ThirdPartyAttributeContent` it uses `expiry_date` so just checking if this is correct.

yoti-sdk-api/src/main/java/com/yoti/api/client/DigitalIdentityClient.java
```diff
@@ -28,11 +31,21 @@ public class DigitalIdentityClient {
...
+    /**
+     * Create a sharing session to initiate a sharing process based on a policy
+     *
+     * @param request
+     *              Details of the request like policy, extensions and push notification for the application
+     * @return an {@link ShareSession}
+     *              Id, status and expiry of the newly created Share Session
+     * @throws DigitalIdentityException
+     *              Aggregate exception signalling issues during the call
+     */
```
pn-santos: This java doc doesn't say much beyond what the method name already says
- `Create a sharing session` -> `createShareSession`
- All params descriptions are essentially saying what could be determined by looking at the API of either `ShareSession` or `ShareSessionRequest`
- `Aggregate exception signalling issues during the call` that's basically a description of _any_ exception, by definition it means "there were issues" that prevented normal execution

I don't have a fancy suggestion for the javadoc here, but usually if I cannot find words to inform beyond what the code already explains then might as well not add any javadoc.

yoti-sdk-api/src/main/java/com/yoti/api/client/DigitalIdentityClient.java
```diff
@@ -34,45 +38,34 @@ public class DigitalIdentityClient {
...
      *
-     * @param request
-     *              Details of the request like policy, extensions and push notification for the application
-     * @return an {@link ShareSession}
-     *              Id, status and expiry of the newly created Share Session
-     * @throws DigitalIdentityException
-     *              Aggregate exception signalling issues during the call
+     * @param request Details of the request like policy, extensions and push notification for the application
+     * @return an {@link ShareSession} ID, status and expiry of the newly created Share Session
+     * @throws DigitalIdentityException Thrown if the session creation is unsuccessful
```
pn-santos: There should be consistency in the formatting of the Javadocs... either always a newline + identation for descriptions or never. But in some places doing it and in others not doesn't look good.

Also FYI, the first sentence of a javadoc should end with a full stop `.`

pn-santos (approved):
only nits/questions

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/extension/ThirdPartyAttributeExtensionBuilder.java
irotech: All good! This builder will create an instance of `ThirdPartyAttributeContent` which defines correctly the properties as

```
    private static final class Property {

        private static final String DEFINITIONS = "definitions";
        private static final String EXPIRY_DATE = "expiry_date";

    }
```

yoti-sdk-api/src/main/java/com/yoti/api/client/DigitalIdentityClient.java
irotech: I was just following what the current "(non)standard" is in place. IMO we could simply not provide any java docs

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/policy/Policy.java
```diff
@@ -0,0 +1,274 @@
...
+        private Builder() {
+            wantedAttributes = new HashMap<>();
+            wantedAuthTypes = new HashSet<>();
+        }
+
+        public Builder withWantedAttribute(WantedAttribute wantedAttribute) {
+            String key = Optional.ofNullable(wantedAttribute.getDerivation()).orElseGet(wantedAttribute::getName);
+
+            if (!wantedAttribute.getConstraints().isEmpty()) {
+                key += "-" + wantedAttribute.getConstraints().hashCode();
```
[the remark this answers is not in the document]
irotech: I could change this but then I should also change the Dynamic Scenario where this code comes from?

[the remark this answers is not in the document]
irotech: I changed the code to use `Object.equals()` instead. The current behavior is in line with the legacy but I'm not completely sure is correct. This requires a chat to double check the logic we want/should enforce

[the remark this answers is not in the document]
irotech: In case of duplicated attributes simply only the first is used

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/policy/Policy.java
```diff
@@ -0,0 +1,282 @@
...
+            wantedAuthTypes = new HashSet<>();
+        }
+
+        public Builder withWantedAttribute(WantedAttribute wanted) {
+            String key = Optional.ofNullable(wanted.getDerivation()).orElseGet(wanted::getName);
+
+            if (wanted.hasConstraints()) {
+                if(wantedAttributes.values().stream()
+                        .map(WantedAttribute::getConstraints)
+                        .noneMatch(c -> c.equals(wanted.getConstraints()))) {
```
pn-santos, line 90: `if(` -> `if (` (space missing)

Also, does the if condition looks readable? maybe extract the condition to a method so it can be _named_

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/policy/Policy.java
```diff
@@ -0,0 +1,282 @@
...
+                    key += "-" + UUID.randomUUID();
+                    wantedAttributes.put(key, wanted);
```
pn-santos, line 92: Unnecessary re-assignment of `key`:
```java
wantedAttributes.put(key + "-" + UUID.randomUUID(), wanted);
```

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/policy/Policy.java
```diff
@@ -0,0 +1,282 @@
...
+                }
+
+                return this;
+            }
+
+            wantedAttributes.put(key, wanted);
+            return this;
```
pn-santos, line 99: Logic of this method seems a bit complicated... wouldn't this result in the same outcome (and be simpler to grok?):
```java
String key = Optional.ofNullable(wanted.getDerivation()).orElseGet(wanted::getName);

if (!wanted.hasConstraints()) {
    wantedAttributes.put(key, wanted);
    return this;
}

if (hasMultipleConstraints(wantedAttributes)) {
    wantedAttributes.put(key + "-" + UUID.randomUUID(), wanted);
}

return this;
```

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/extension/ThirdPartyAttributeExtensionBuilder.java
pn-santos: Maybe I didn't explain correctly, I know that `ThirdPartyAttributeContent` has `expiry_date`, what I'm asking (assuming that is correct) is why _this_ class (`ThirdPartyAttributeExtensionBuilder`) has `expiryDate` instead of `expiry_date`.... I guess these are not JSON properties but _fields_ and are used only for exception messages?

Even so, why the mismatch if they both refer to the same thing? or do they refer to different things?

irotech: gonna change to match the JSON property. As you said this field is used for the exception msg

yoti-sdk-api/src/main/java/com/yoti/api/client/identity/policy/Policy.java
```diff
@@ -0,0 +1,285 @@
...
+
+        public Builder withWantedAttribute(WantedAttribute wanted) {
+            String key = Optional.ofNullable(wanted.getDerivation()).orElseGet(wanted::getName);
+
+            if (!wanted.hasConstraints()) {
+                wantedAttributes.put(key, wanted);
+                return this;
+            }
+
+            if (constraintsNotPresent(wanted.getConstraints())) {
```
pn-santos, line 92: `hasNoConstraints` (it's preferable, if feasible, for boolean returning methods to use `is` or `has` type prefixes).

Also AFAICT this method is checking whether there's constraints in `wantedAttributes` that are not equal to those in `wanted` right? (`true` if there aren't, `false` if there are)

So maybe this method would be clearer if it was something like:
```java
if (hasIdenticalConstraints(wantedAttributes. wanted))
```
(turning it into a static method)

pn-santos (approved):

pn-santos (approved):

PR #436, opened 2024-03-12
Update issue templates

irotech:


irotech:
After discussing with integration we only need to update the support email to converge all the support requests, from all the SDKs, into the same bucket. Work has been done in [PR-433](https://github.com/getyoti/yoti-java-sdk/pull/433)

PR #437, opened 2024-03-13
SDK-2431: Drop null check for optional receipt properties

irotech:


yoti-sdk-api/src/main/java/com/yoti/api/client/ActivityDetails.java
```diff
@@ -1,7 +1,6 @@
 package com.yoti.api.client;
 
-import static com.yoti.api.client.spi.remote.util.Validation.notNull;
-
+import static com.yoti.validation.Validation.notNull;
```
pn-santos, line 3: is this right? the diff doesn't show any _new_ `notNull` calls so I don't see why suddenly this import is needed...

yoti-sdk-api/src/main/java/com/yoti/api/client/ActivityDetails.java
```diff
@@ -13,26 +12,26 @@ public final class ActivityDetails {
...
             ExtraData extraData,
             Date timestamp,
             byte[] receiptId) {
-        this.rememberMeId = notNull(rememberMeId, "Remember Me id");
+        this.rememberMeId = rememberMeId;
         this.parentRememberMeId = parentRememberMeId;
-        this.userProfile = notNull(userProfile, "User profile");
-        this.applicationProfile = notNull(applicationProfile, "Application profile");
+        this.otherPartyProfile = otherPartyProfile;
+        this.thisPartyProfile = thisPartyProfile;
```
pn-santos, line 31: I could be missing something, but `ActivityDetailsFactory` is always passing non null instances of `HumanProfile` and `ApplicationDataProfile`:
```java
public ActivityDetails create(Receipt receipt, PrivateKey privateKey) throws ProfileException {
    (...)
    
    HumanProfile otherPartyProfile = new HumanProfile(otherPartyAttr);
    ApplicationProfile thisPartyProfile = new ApplicationProfile(thisPartyAttr);
    
    (...)
    
    return new ActivityDetails(
            (...)
            otherPartyProfile,
            thisPartyProfile,
           (...)
    );
}
```

So while the _content_ of those instances might be `null` (though I would expect them to be empty lists instead of `null` but I didn't check the codebase to be sure) the instances themselves (of `HumanProfile` and `ApplicationDataProfile`) are never `null`.

Whether or not the checks should be kept, that's another question, but personally if it's being done for other fields why not these if they are indeed never `null`?

irotech: I triple checked and these should never be `null` (if not for a bug) I'll put the check back. Good spot.

yoti-sdk-api/src/main/java/com/yoti/api/client/ActivityDetails.java
pn-santos: ah ok - ideally if that import change had been done in it's own commit it would have been clearer 😉

pn-santos (approved):

PR #438, opened 2024-03-14
SDK-2364: Map attribute alternative names

irotech:
The [epic](https://lampkicking.atlassian.net/browse/SDK-2360) suggests to map both `anchors: []` and `alternative_names: []` but turns out the `Attribute.anchors` has been [deprecated](https://github.com/lampkicking/core-api/blob/7da190ab1e4eb2b8c29fe973e80b91f015e91278/sharepubapi_v1/Policy.proto#L129) for years (see [conversation](https://yoti.slack.com/archives/C3B8FSJMV/p1710427120485889) for details)

### DIFF
--- a/yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/WantedAttribute.java
+++ b/yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/WantedAttribute.java
@@ -4,49 +4,48 @@ import static com.yoti.validation.Validation.notNullOrEmpty;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import com.yoti.api.client.shareurl.constraint.Constraint;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 /**
- * Type and content of an user detail
+ * Type and content of a user detail
  */
 public class WantedAttribute {
 
-    @JsonProperty("name")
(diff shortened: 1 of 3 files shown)


yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/WantedAttribute.java
```diff
@@ -135,10 +152,35 @@ public Builder withConstraint(Constraint constraint) {
             return this;
         }
 
+        public Builder withAlternativeNames(Set<String> alternativeNames) {
+            this.alternativeNames = Collections.unmodifiableSet(alternativeNames);
```
pn-santos, line 156: I realise this was being done, but using `Collections.unmodifiableSet` here doesn't make sense, it _doesn't_ help make the set of strings immutable, as the caller can still add/remove elements from the set that was passed in (causing the state of the Builder's `alternativeNames` to change.

If the goal is to enforce immutability, then the set needs to be copied:
```java
this.alternativeNames.addAll(alternativeNames`)
```
and then make the copy immutable 
```java
this.alternativeNames = Collections.unmodifiableSet(builder.alternativeNames);
```
(typically it's preferable to leave the builder set mutable because there might be a need in the future to have setters for single elements that just _add_ to an existing set rather than complete replacement).

same for any other collection where immutability needs to be enforced.

irotech:
I added and extra commit to change v2 as well, and rebased

pn-santos (approved):

PR #439, opened 2024-03-18
SDK-2367: Add support for advanced identity profiles to Share V1

irotech:


yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/DynamicPolicy.java
```diff
@@ -94,6 +94,15 @@ public Object getIdentityProfile() {
         return identityProfile;
     }
 
+    /**
+     * Defines a required advanced identity profile.
+     *
+     * @return AdvancedIdentityProfile
+     */
```
pn-santos, line 101: These type of javadocs are useless... it is basically just parroting what the method name already says.

Also the summary doesn't even make sense tbh "**Defines** a required...", it's a `get` how does it _define_ anything?

Drop it. Or re-write it so that it provides _more_ information that isn't already embedded in the method name.

pn-santos (approved):

yoti-sdk-api/src/main/java/com/yoti/api/client/shareurl/policy/DynamicPolicy.java
irotech: Going to drop all of them. We do not provide any for `v2` and I do agree the value there is close to 0

pn-santos (approved):

irotech:
@pn-santos Added README update within the rebase

pn-santos (approved):

PR #455, opened 2024-09-20
SDK-2471: Add support for brand ID to be set at sdk config creation

artem-yoti:
Here we are adding support for brand ID to be set at sdk config creation.


artem-yoti:
Еt turned out that this change was already done a month ago - https://github.com/getyoti/yoti-java-sdk/pull/452
I'm closing this PR

PR #458, opened 2024-09-23
SDK-2340: Allow Relying Business to configure if expanded document fi…

MrBurtyyy:
…elds should be created for a text-extraction task


bucky-boy (approved):
I've amended some Java Doc comments and added a missing UT @MrBurtyyy , please take a look before merging

PR #488, opened 2025-07-16
SDK-2640: Migrate to Sonatype Maven Central

irotech:


pn-santos (commented):
Don't you also need to change the `distributionManagement` ? It's currently
```xml
<distributionManagement>
  <snapshotRepository>
    <id>ossrh</id>
    <url>https://oss.sonatype.org/content/repositories/snapshots</url>
  </snapshotRepository>
  <repository>
    <id>ossrh</id>
    <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
  </repository>
</distributionManagement>
```
which doesn't seem right given that the ossrh was discontinued.

irotech:
> Don't you also need to change the `distributionManagement` ? It's currently
> 
> ```
> <distributionManagement>
>   <snapshotRepository>
>     <id>ossrh</id>
>     <url>https://oss.sonatype.org/content/repositories/snapshots</url>
>   </snapshotRepository>
>   <repository>
>     <id>ossrh</id>
>     <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
>   </repository>
> </distributionManagement>
> ```
> 
> which doesn't seem right given that the ossrh was discontinued.

Good question. The repository is no longer required and we do not deploy snapshots so I'm checking if we could drop the entire `distributionManagement` all together

pn-santos:
A cursory look in github shows this:
```xml
<distributionManagement>
  <repository>
    <id>central</id>
    <url>https://central.sonatype.com</url>
  </repository>
  <snapshotRepository>
    <id>central-snapshot</id>
    <url>https://central.sonatype.com/repository/maven-snapshots/</url>
  </snapshotRepository>
</distributionManagement>
```
If we don't deploy snapshots, then I guess only the `repository` would apply. 

> The repository is no longer required

Pretty sure `distributionManagement` is used by the `maven-deploy-plugin` so if now we no longer use it then it's not needed. That said not 100% it might not have other uses.

Like I said, a brief look around other OSS projects that did the conversion, most seem to keep `distributionManagement` element with the above config (but perhaps it's done for documentation purposes?)

MrBurtyyy (approved):

